PR #7, opened 2017-03-21
Draw enhancements

igorti:
This PR adds some tweaks to the Draw component. 

Before this can be merged one issue should be solved - when user closes draw dialog(and features remain on the map) and the clicks on the drawn feature, the dialog gets opened, but then Identify tool remains enabled. I couldn't figure out how to disable it for this scenario, so any suggestions appreciated.

### DIFF
--- a/QWC2Components/actions/task.js
+++ b/QWC2Components/actions/task.js
@@ -14,7 +14,6 @@ const {changeMapInfoState} = require('../../MapStore2/web/client/actions/mapInfo
 function setCurrentTask(task, mode=null, allowIdentify=false) {
     return (dispatch) => {
         dispatch(changeMeasurement({geomType: null}));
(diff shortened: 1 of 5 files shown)


manisandro:
It depends on how you re-activate the draw mode. You should switch tasks by calling the `setCurrentTask` action (https://github.com/qgis/qwc2/blob/master/QWC2Components/actions/task.js#L14). This then disables identify unless `allowIdentify` is `true`.

manisandro:
How did you update your branch? My recent commits shouldn't be part of the PR.

igorti:
Sorry, didn't notice that I accidentally rebased this branch. Should be fixed now. 

Don't merge yet as I'm still looking for a proper way to call `setCurrentTask`

PR #8, opened 2017-04-11
Draw update and translation

karl-magnus:
Draw color picker toggling and translation of draw styling elements


manisandro:
Thanks

PR #10, opened 2017-05-23
fix loop in print grid

luciman:
fixes https://github.com/qgis/qwc2-demo-app/issues/32


tudorbarascu:
Just tested and it fixes the Printing issues.
thanks @luciman

manisandro:
Thanks!

PR #11, opened 2017-05-26
fix tree issue for group toggle sublayers

luciman:
Hi,

When Upper group is unchecked, toggling on subgroup/sublayer does
nothing. This fixes it.
E.g. If you uncheck the root tree, although you toggle on any
sublayer/subgroup, there's no image showing up.

coco


tudorbarascu:
@luciman Thanks a lot!

manisandro:
Thanks

PR #12, opened 2017-05-30
Fix GetFeatureInfo in duplicate projects

tudorbarascu:
Hello,

If the same project is used (multiple themes pointing to it),
the old behavior added an id to the name of the layer which resulted
in GetFeatureInfo requests being sent to a false layer name consisting
of real_lyrname + id (0 in my case).

### DIFF
--- a/scripts/themesConfig.js
+++ b/scripts/themesConfig.js
@@ -290,7 +290,7 @@ function getTheme(configItem, resultItem) {
 
             // update theme config
             resultItem.id = themeId;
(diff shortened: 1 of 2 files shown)


luciman:
Thank you,
It would be nice if you'll change also in themesConfig.py because some of us use Windows

tudorbarascu:
@luciman Thanks, I've pushed another commit that should do it

manisandro:
Thanks

PR #13, opened 2017-05-31
Enhancement: don't display background lyr switcher when no back lyrs exist

luciman:
The background layer switcher shouldn't be displayed when there
aren't any background layers.

### DIFF
--- a/QWC2Components/plugins/BackgroundSwitcher.jsx
+++ b/QWC2Components/plugins/BackgroundSwitcher.jsx
@@ -32,18 +32,21 @@ const BackgroundSwitcher = React.createClass({
     },
     render() {
         let activeClass = this.props.visible ? 'active' : '';
-        return (
-            <div>
-                <Button id="BackgroundSwitcherBtn" className={activeClass}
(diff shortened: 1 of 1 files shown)


QWC2Components/plugins/BackgroundSwitcher.jsx
```diff
@@ -32,18 +32,21 @@ const BackgroundSwitcher = React.createClass({
...
+                    <div id="BackgroundSwitcher" className={activeClass}>
+                        {this.renderLayerItem(null, this.props.layers.filter(layer => layer.visibility === true).length === 0)}
+                        {this.props.layers.map(layer => this.renderLayerItem(layer, layer.visibility === true))}
+                    </div>
                 </div>
-            </div>
-        );
+            );
+        }
+        return (<div></div>);
```
manisandro: You should also be able to return just `null`.

luciman: Indeed. For some reason I thought it wouldn't work (Had the feeling I had tried it in the past without it working).

Rebased now. Thanks a lot!_

manisandro:
Thanks

PR #14, opened 2017-05-31
get default theme by ID

luciman:
Before
https://github.com/qgis/qwc2/commit/f7559eb720e4f68c62ecb66af31153580f2c3e50
the themeId was the same as themeName and it all worked fine.

This fixes the Default theme for the case where themes have the same
root.


manisandro:
Thanks

PR #15, opened 2017-06-15
themesconfig.js and package-commands.sh writes bad config

huggla:
The exit code of **package-commands.sh** is always 0 (success). **themesconfig.js** detects a bad **themesConfig.json** and outputs error but **themes.json** is still overwritten and the exit code is 0 (success).

¤ A good **themes.json** must not be overwritten with a bad **themes.json**.
¤ **themesconfig.js** should always fail if **themesConfig.json** contains errors.
¤ Scripts should always return correct exit code.


manisandro:
Thanks!

PR #16, opened 2017-07-12
Local Coordinate system support

mohammadamin46bd:
Local Coordinate system support:
Theme specific WMS version for supporting local coordinate system


manisandro:
Also here, what's the goal of setting the center globally? It is really a theme specific setting, controllable through the theme extent.

mohammadamin46bd:
Ok, setting center globally is user-specific and will remove it. I have checked there are some problem with theme extent for different crs but might not be related for this PR.

I will update this PR as wms version 1.3.0 has some problem for different crs and it might be good to set the wms version from the configuration.

manisandro:
Yes agree with the WMS version - though question again is whether it should be theme specific, since the used WMS server is theme specific.

mohammadamin46bd:
Is it not better to read the theme specific WMS version from the configuration? Now the client is always requesting WMS version 1.3.0 and some WMS server fails to response.

So, if a user set theme specific WMS version it will request with the theme specific WMS version otherwise it will request as it is now on the client (WMS version 1.3.0).

manisandro:
Uhm not following your argument, you say that you'd rather set a specific WMS version globally and override as necessary instead of assuming 1.3.0 by default and override as necessary in the themes?

If you want to change the global default, I'd rather introduce a `defaultWMSVersion` similar to [1] in addition to the theme specific `WMSVersion` setting - it makes the logic simpler since themeConfig.py / themeConfig.js will take care of setting the right version in the generated themes.json, and then you just need to tweak ThemeSwitcher::createLayerForTheme [2] to also read the WMS version from the theme config.

[1] https://github.com/qgis/qwc2-demo-app/blob/master/themesConfig.json#L304
[2] https://github.com/qgis/qwc2/blob/2327622b4f9ed4b16cac7fc2b211deed4687ca0c/QWC2Components/plugins/ThemeSwitcher.jsx#L185

manisandro:
Could you clean up the PR and add just what is relevant? (Is there more than the WMS version logic?)

mohammadamin46bd:
Yes, it is now only about WMS version logic in general.

manisandro:
Superseded by #18.

PR #17, opened 2017-07-13
Handle relative urls in themes config

huggla:
Fix themesConfig to work with relative urls in themesConfig.json.

### DIFF
--- a/scripts/themesConfig.js
+++ b/scripts/themesConfig.js
@@ -13,6 +13,7 @@ const axios = require('axios');
 const xml2js = require('xml2js');
 const fs = require('fs');
 const fqdn = require('node-fqdn');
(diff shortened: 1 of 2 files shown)


manisandro:
qwc2serverUrl is actually only meant to specify the optional QWC2 server-side component which does the permalink storage/resolution and possibly proxying (to avoid CORS issues). It is not the QGIS Server URL.

huggla:
Sorry, I got a bit confused. I've updated my patch of themesConfig.js and themesConfig.py.

manisandro:
So this basically attempts to detect the FQDN of the current machine and use that as baseurl?

Can you squash the commits as appropriate?

huggla:
I'm having 4 weeks vacation. I'll do it when I'm back at work.

PR #19, opened 2017-08-22
Fix themesConfig to work with relative urls.

huggla:
Attempts to detect the FQDN of the current machine and use that as baseurl.

### DIFF
--- a/scripts/themesConfig.js
+++ b/scripts/themesConfig.js
@@ -12,6 +12,8 @@ const urlUtil = require('url');
 const axios = require('axios');
 const xml2js = require('xml2js');
 const fs = require('fs');
+const fqdn = require('node-fqdn');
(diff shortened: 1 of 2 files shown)


manisandro:
So the goal here is that in themesConfig.json you can write qgis server WMS urls without a hostname, and which will then be resolved to the current hostname? I take an application of this is to be able to deploy the themesConfig to multiple machines running their own qgis server instance?

huggla:
1. We run our services as docker containers, they need to be clone-able.

2. Currently, having relative urls in themes.json works fine, but the themesConfig script resolves to host IP address instead of FQDN. Our load balancer doesn't like IP addresses and we get "ERROR reading WMS GetProjectSettings". We don't want to edit themes.json manually.

manisandro:
Okay thanks!

PR #20, opened 2017-08-30
Fix the possibility of users setting rotation to Blank

luciman:
It is currently possible to delete the rotation angle, thus
setting the angle to NaN value. which in turn outputs a blank map.
This fixes it.


manisandro:
Thanks!

PR #25, opened 2017-09-11
Enhance the search so that marker stays on feature

luciman:
Fixes https://github.com/qgis/qwc2-demo-app/issues/23

### DIFF
--- a/QWC2Components/components/Search.jsx
+++ b/QWC2Components/components/Search.jsx
@@ -327,23 +327,22 @@ const Search = React.createClass({
                 this.props.mapConfig.projection);
         }
         if(item.provider && this.props.searchProviders[item.provider].getResultGeometry) {
-            this.props.searchProviders[item.provider].getResultGeometry(item, this.showFeatureGeometry);
-        }
-        this.setState({currentResult: item});
-    },
-    showFeatureGeometry(item, geometry, crs) {
-        if(item === this.state.currentResult) {
-            let feature = IdentifyUtils.wktToGeoJSON(geometry);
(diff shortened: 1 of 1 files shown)


QWC2Components/components/Search.jsx
```diff
@@ -336,8 +335,44 @@ const Search = React.createClass({
         if(item === this.state.currentResult) {
             let feature = IdentifyUtils.wktToGeoJSON(geometry);
             feature.geometry = IdentifyUtils.reprojectFeatureGeometry(feature.geometry, crs, this.props.mapConfig.projection);
+            let geojson  = new ol.format.GeoJSON().readFeature(feature)
+            let center = this.getFeatureCenter(geojson.getGeometry())
+            let wgscenterlatlon = CoordinatesUtils.reproject(center, item.crs, "EPSG:4326");
+            let text = item.label !== undefined ? item.label : item.text;
```
manisandro: Isn't this like a duplicate with line 297?

luciman:
```let text = item.label !== undefined ? item.label : item.text;``` is duplicate.
Indeed. The problem is that I'm a reactjs newbie and I don't know how to pass the text variable to showFeatureGeometry() I'll refactor this once I learn how to do it. Any hints are appreciated.
Thanks.

manisandro:
You can just use an inline function 

    (item, geometry, crs) => { showFeatureGeometry(item, geometry, crs, label) }

but in this case, just remove the first `let text [...]` line, it's unused AFAICS.

luciman:
@manisandro Thanks for the advice. It's reworked now. I also had an error for the marker on coordinate search.

manisandro:
Thanks!

PR #26, opened 2017-09-12
improve features highlighting in identify viewer

luciman:
highlight just the features from layer and NOT all the
results when "mouse over"ing over layer label

### DIFF
--- a/QWC2Components/components/IdentifyViewer.jsx
+++ b/QWC2Components/components/IdentifyViewer.jsx
@@ -107,7 +107,11 @@ const IdentifyViewer = React.createClass({
     setHighlightedFeatures(feature, resultTree) {
         let features = [];
         if(feature) {
-            features = [feature];
+            if(feature instanceof Array){
+                features = feature;
+            } else{
+                features = [feature];
+            }
         } else {
             Object.keys(resultTree).map(key => {
                 features = features.concat(resultTree[key].map(feature => assign({}, feature, {id: key + "." + feature.id})))}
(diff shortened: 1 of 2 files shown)


QWC2Components/components/IdentifyViewer.jsx
```diff
@@ -107,7 +107,11 @@ const IdentifyViewer = React.createClass({
     setHighlightedFeatures(feature, resultTree) {
         let features = [];
         if(feature) {
-            features = [feature];
```
manisandro: I think it would be more elegant to just change `setHighlightedFeatures(feature, resultTree)` to `setHighlightedFeatures(features, resultTree)` ( so feature**s** plural) and ensure `setHighlightedFeatures` always receives an array from the caller.

luciman:
@manisandro  How does this looks now?

manisandro:
Looks good to me, thanks!

manisandro:
@luciman There actually seems to be a regression here: if you don't just hover on a feature entry in the identify results, but actually click on it, other features are selected as well. Can you reporduce this?

luciman:
@manisandro I try to reproduce your issue but I'm not sure I understand it.
If you click on a feature that feature is highlight on the map. 
Even if the attribute-list-box of that clicked (selected) feature is open  if you hover with mouse other feature, on the map that hover feature is highlight. 
If you go with the mouse on map (outside the info window) just the selected (clicked) feature remains highlight. 
It can't be selected two features same time.
The issue you describe is after my commit? or even before was the same behavior?

manisandro:
See video [1]: at approx 00:05, when I click on `202`, all features are highlighted, not just the one I clicked.

Looks like this behaviour started after 9bb8cb4e445d9d14a17b0700ba5359bea4ce9827.

[1] https://smani.fedorapeople.org/tmp/out-1.ogv

PR #27, opened 2017-09-12
untoggle feature if the parent layer is colapsed/untoggled

luciman:
If you select a feature and you collapse the parent layer the feature
remains selected. The normal behavior in my mind is that feature should
be deselected if the user untoggles the parent layer.

What do you think?


manisandro:
Yeah this makes sense, thanks!

PR #29, opened 2017-09-15
fix z-index of resizable-window

luciman:
before this, the info window could be moved under the top bar, nulling the possibility of easy closing.
before

after
after


manisandro:
@luciman Did you check whether it behaves correctly with the appmenu?

luciman:
Yes it is ok. it is over LayerTree menu. The appmenu closes automatically if you move the window as before

manisandro:
Ok thanks!

PR #31, opened 2017-09-18
configurable shareplugin via config.json

stephaschu:
for example, if the webclient runs on intranet there is no need for sharing it in social media.

config.json:

        {
          "name": "Share",
          "cfg": {
            "showSocials": false,
            "showLink": true,
            "showQRCode": true
          }
        }


manisandro:
Thanks!

PR #35, opened 2017-11-23
set parent bounds for resizable window

luciman:
fix - if the user drag the identify window up outside the map window, there is no possibility to drag it back or close

screenshot from 2017-11-23 12-53-09


manisandro:
Ah cool!, thanks!

PR #42, opened 2018-01-11
Fix: Uncaught (in promise) ReferenceError: ol is not defined

stephaschu:
Uncaught (in promise) ReferenceError: ol is not defined
at Search._this.showFeatureGeometry (Search.jsx:213)
at eval (Search.jsx:203)
at eval (SearchProviders.js:71)


manisandro:
Thanks!

PR #46, opened 2018-02-21
fix Ios mobile AppMenu Click

luciman:


manisandro:
Thanks - just to understand, what happens currently?

luciman:
on iphone Safari or chrome when tap/click on burger icon for open the AppMenu, nothing happens.

manisandro:
Uhm ok, but does `checkCloseMenu` method get called? Unless the `checkCloseMenu` is misbehaving, I wonder what killing the click event actually does?

luciman:
I think on iphone without  ev.preventDefault(); and  ev.stopPropagation(); is like you'll click on div className="appmenu-button-container" and on div id=TopBar 
Try on desktop to click on the icon and after the menu will be opened click on TopBar blank zone (the left fone of the menu). In my opinion that is what is happening on iphone.

luciman:
the menu will open and close immediately even with `checkCloseMenu`

manisandro:
So for instance when clicking a submenu, can you trace who actually calls `toggleMenu`? Cause that should be the only way a menu can be closed again. It should be either because `appmenu-button-container` receives a mousedown, in which case killing the click should actually make any difference, or because `checkCloseMenu` is triggered and `!this.menuEl.contains(ev.target)` is mis-evaluated as true.

luciman:
It is very hard for me to debug on an iphone.
The behaviour is like:
without  `onClick={this.killEvent}` in `id="AppMenu"` the menu is not opening
without `onClick={this.killEvent}` in `#AppMenu ul.appmenu-menu li`, `this.onMenuitemClicked` is not working.
what is very strange is that `this.onMenuitemClicked` is working if it is touch `#AppMenu ul.appmenu-menu li img`

manisandro:
I'll try and find someone with such a device and see if I can investigate.

manisandro:
So this appears to be https://github.com/facebook/react/issues/134  - ugh...

luciman:
With that killEvent it works. I tested. I don't know if it is the best solution but it works

manisandro:
It also works to just have an empty `onClick={ev => {}}` on `<div id="AppMenu">`. Trying to find out whether any other places are affected.

manisandro:
I've addressed this with https://github.com/qgis/qwc2/commit/ad2f779fe4aa78ed2f08c0163b94b00508b909a0.

PR #47, opened 2018-02-25
Fix broken MapTip

stephaschu:
When you click on "Show layer map tips" an error is thrown:
> Uncaught TypeError: _this.props.toggleMapTips is not a function


manisandro:
Thanks!

PR #51, opened 2018-03-22
allow config decimal precision for HeightProfile MaoInfo Tooltip

luciman:
https://github.com/qgis/qwc2-demo-app/issues/82

### DIFF
--- a/QWC2Components/plugins/HeightProfile.jsx
+++ b/QWC2Components/plugins/HeightProfile.jsx
@@ -27,10 +27,12 @@ class HeightProfile extends React.Component {
         samples: PropTypes.number,
         mobile: PropTypes.bool,
         addMarker: PropTypes.func,
-        removeMarker: PropTypes.func
+        removeMarker: PropTypes.func,
+        heighProfilePrecision: PropTypes.number
     }
     static defaultProps = {
-        samples: 500
+        samples: 500,
+        heighProfilePrecision: 0
     }
     static contextTypes = {
         messages: PropTypes.object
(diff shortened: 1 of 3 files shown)


manisandro:
Thanks! Since the respective props are used just by a single component, I'd think it would be cleaner to declare these as props in the components and to configure them in cfg of the respective plugin in `config.json`.

QWC2Components/plugins/HeightProfile.jsx
```diff
@@ -136,10 +136,11 @@ class HeightProfile extends React.Component {
                         this.updateMapMarker(idx / this.props.samples * totLength);
                         if(this.tooltip) {
                             let sample = data.series[0][idx];
+                            let heighProfilePrecision = ConfigUtils.getConfigProp("heighProfilePrecision") || 0;
                             this.marker.style.visibility = this.tooltip.style.visibility = 'visible';
                             this.marker.style.left = this.tooltip.style.left = ev2.clientX + 'px';
-                            this.tooltip.innerHTML = "<b>" + distanceStr + ":</b> " + Math.round(sample.x) + " m<br />" +
-                                                     "<b>" + heightStr + ":</b> " + Math.round(sample.y) + " m " + aslStr;
+                            this.tooltip.innerHTML = "<b>" + distanceStr + ":</b> " + Math.round(sample.x * Math.pow(10, heighProfilePrecision))/Math.pow(10, heighProfilePrecision) + " m<br />" +
+                                                     "<b>" + heightStr + ":</b> " + Math.round(sample.y * Math.pow(10, heighProfilePrecision))/Math.pow(10, heighProfilePrecision) + " m " + aslStr;
```
manisandro: You might also want to introduce a variable for the `Math.pow(10, heighProfilePrecision)` result and then use that one, in the interest of code readability.

luciman:
@manisandro It should be done now as you suggested. Thanks for reviewing!

manisandro:
Thanks!

PR #53, opened 2018-04-11
fix typo in DxfExport

stephaschu:
parameter FILENAME should be FILE_NAME (see https://docs.qgis.org/2.18/en/docs/user_manual/working_with_ogc/ogc_server_support.html#dxf-export)


manisandro:
Thanks!

PR #56, opened 2018-06-04
apply tiled-parameter from qwc2_demo_app

tobias-brunner:
should fix:

https://github.com/qgis/qwc2-demo-app/issues/94


manisandro:
Thanks!

PR #58, opened 2018-06-13
Russian localization 

novikov-alexander:
I tried to translate not straightforward but maximizing the meaning and stylistical correctness


manisandro:
Thanks - the change in LocaleUtils is not necessary as the supported locales are overridden by those set in js/appConfig.js anyways, so just add the locale specification there. Actually, I should remove the list in LocaleUtils.js to avoid confusion.

novikov-alexander:
@manisandro I didn't know it.
Thanks =)

manisandro:
Committed in e336782aa70f0411e86509a2242e5936f216c2ec, thanks!

PR #60, opened 2018-07-03
Leep layer params on layer adding and updating

novikov-alexander:
I suggest to keep layer params while it saving to store. It may be useful if we want "FILTER" to be added to params object.


manisandro:
Looks good to me, thanks!

PR #61, opened 2018-07-10
remove html tags fom search text for marker label

luciman:
for example http://api3.geo.admin.ch/rest/services/api/SearchServer returns label with html tags and the ol.style.text display it like string


manisandro:
Thanks!

PR #62, opened 2018-07-10
scroll feature details in resizable window

luciman:


manisandro:
Why is this needed? The parent element already has `overflow: auto`.

luciman:
go on Liegenschaftsprojekte Stadt Uster  theme and click on SH Gschwader

by default the feature window is like
feature info1
without scroll and for view all the details you have to resize the window like
feature info2
if the screen have the resolution 1366x768 there will be a problem

manisandro:
Ah yes indeed, something broke there. Addressed in ae1cac9643c386d2a0f498ad7e67b0c973505106 (fix is slightly more complicated than just adding overflow, as to ensure it works correctly both in tree and flat display modes of the identify results). Thanks for reporting!

PR #63, opened 2018-09-19
Redundant offset while clicking positioned map fixed

novikov-alexander:
If map container is positioned via "top","left","margin","padding" etc. maptiptool opens in a wrong place and identifytool identifies wrong coordinates. 

I've solved this problem.


manisandro:
Thanks!

PR #64, opened 2018-11-22
Czech translation

freedom7979:
I translated qwc2 into Czech. Please commit is it possible.
Thanks


manisandro:
Thanks!

freedom7979:
Thanks as well. I have a question. Is this location "qwc2-demo-app/translations/" actual? Do I have to do commit this loaction? Thanks you for response.

freedom7979:
Thank you for the explanation. So it means no action for me.

PR #65, opened 2018-12-07
ensure rows is an array

stephaschu:
otherwise rows.concat line 281 fails with: rows.concat is not a function


manisandro:
Thanks!

PR #70, opened 2019-02-13
prevent not visible highlight-layers from being printed

tobias-brunner:
currently highlight-layers are always printed (getPrint).


manisandro:
Thanks!

PR #72, opened 2019-02-21
close searchresults upon adding layer to layertree

tobias-brunner:
close searchresults upon adding layer to layertree.

Not sure if this was intended, though...


manisandro:
Thanks!

PR #73, opened 2019-03-11
Allow to pass http basic credentials during configuration

autra:
This is very useful when building the sites on dev machine or using an external maps service.

Note that the latter use case is still not completely supported by qwc2, because some runtime requests like getFeatureInfo needs proper CORS configuration if the front has a different Origin than the map server.

Supporting the latter will require being able to configure axios on a per-host basis.

This implements my request on [qwc2-example-app repo](https://github.com/qgis/qwc2-demo-app/issues/141).

### DIFF
--- a/scripts/themesConfig.js
+++ b/scripts/themesConfig.js
@@ -87,7 +87,10 @@ function getThumbnail(configItem, resultItem, layers, crs, extent, resolve) {
     parsedUrl.query.LAYERS = layers.join(',');
     const getMapUrl = urlUtil.format(parsedUrl);
 
-    axios.get(getMapUrl, {responseType: "arraybuffer"}).then((response) => {
+    axios.get(getMapUrl, {
+        responseType: "arraybuffer",
+        auth: configItem.wmsBasicAuth,
+    }).then((response) => {
         let basename = configItem.url.replace(/.*\//, "").replace(/\?.*$/, "") + ".png";
         try {
             fs.mkdirSync("./assets/img/genmapthumbs/");
@@ -222,7 +225,7 @@ function getTheme(config, configItem, result, resultItem) {
     const getCapabilitiesUrl = urlUtil.format(parsedUrl);
 
     return new Promise((resolve, reject) => {
-        axios.get(getCapabilitiesUrl).then((response) => {
+        axios.get(getCapabilitiesUrl, { auth: configItem.wmsBasicAuth, }).then((response) => {
             // parse capabilities
             let capabilities;
             xml2js.parseString(response.data, {
--- a/scripts/themesConfig.py
+++ b/scripts/themesConfig.py
@@ -8,9 +8,9 @@
 
 import os
 try:
-    from urllib.request import urlopen
+    from urllib import request
 except:
-    from urllib2 import urlopen
+    import urllib2 as request
 try:
     from urllib.parse import quote, urljoin
 except:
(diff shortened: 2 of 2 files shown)


scripts/themesConfig.js
```diff
@@ -22,6 +22,14 @@ try {
 const uuid = require('uuid');
 
 const hostFqdn = fqdn ? "http://" + String(fqdn()) : "";
+const auth = process.env["QWC2_WMS_HTTP_AUTH"] || null;
```
elemoine: You don't need `|| null`, do you?

elemoine:
Just a very minor comment, otherwise looks good to me.

It would be good to document the new envvar. With a patch to qwc2-demo-app probably?

scripts/themesConfig.js
autra: Not really indeed. A weird reflex of always initializing stuff and because I tend to consider there's a semantic difference between `null` and `undefined`... but it's a bit useless here and it works only if this convention is followed everywhere in the project.

elemoine: Agree :100:%

manisandro:
Thanks! - Would be nice to also have this in the python variant, to keep them from diverging.

autra:
> Would be nice to also have this in the python variant, to keep them from diverging.

I'm currently doing that. 

I'm not a python expert, but it seems to me that python 2.7 support is intended (from [this part](https://github.com/qgis/qwc2/blob/master/scripts/themesConfig.py#L10)). However, on my machine, the script fails [here](https://github.com/qgis/qwc2/blob/master/scripts/themesConfig.py#L433) (`TypeError: 'encoding' is an invalid keyword argument for this function`). Is python 2.7 support broken, or am I wrong making this assumption?

elemoine:
> I'm not a python expert, but it seems to me that python 2.7 support is intended (from [this part](https://github.com/qgis/qwc2/blob/master/scripts/themesConfig.py#L10)). However, on my machine, the script fails [here](https://github.com/qgis/qwc2/blob/master/scripts/themesConfig.py#L433) (`TypeError: 'encoding' is an invalid keyword argument for this function`). Is python 2.7 support broken, or am I wrong making this assumption?

Maybe, you can make it work for Python 3 and we will fix it for Python 2.7 later with a separate PR. @manisandro, do you agree?

manisandro:
Yes, agree

autra:
After trying to do the python version, I realised that we actually don't want to send these credentials to *all* servers defined in themesConfig.json, but rather send only this authent to the relevant server. We don't want to send credentials to everybody!

So we actually need a list of (url, user, pass) triplet. I see several possibilities:
1. have a more complex format for the env var. 
2. have this list of triplets somewhere in themesConfig.json
3. or have the user / pass directly in a theme item in themesConfig.json


N°1 seems impractical to me, because whatever separator we choose, we would need to give the user the possibility to escape it. For instance,  we can reasonably expect user not to contain `:` (and this is already a tradeoff), but not so much with password, so if we have several in the string, it can cause issues.

N°2 and 3 would work. N°3 has the advantage of keeping the information near where it is relevant. What do you think @elemoine @manisandro ?

elemoine:
@autra very good point! In that case my vote goes to option 3 and a new theme item property: `wmsBasicAuth`. We will need to document it, and clearly state that this property is only used by the themesConfig script.

manisandro:
Yes agree for option 3.

autra:
@manisandro I've updated this PR and opened a PR for the documentation: https://github.com/qgis/qwc2-demo-app/pull/142

pinging @elemoine too.

Don't hesitate to read carefully my python code :-)

manisandro:
Thanks!

elemoine:
Great! Thanks @autra and @manisandro.

autra:
thanks!

PR #74, opened 2019-03-18
override img vertical-align for icons

stephaschu:
otherwise img icons in TopBar are not inline:

before:
topbar
after:
topbar2


manisandro:
Thanks!

PR #78, opened 2019-06-08
FeatureInfo on external WMS Layers added with ImportLayer

stephaschu:
FeatureInfo on external WMS Layers added with ImportLayer get executed but the IdentifyViewer says "No information available for the selected point". Problem is that the request is called with info_format=text/xml as WMS servers responding with a different content which is not getting parsed by IdentifyUtils.

So i did a little fix here to prefer "text/html" for ServiceLayerUtils: https://github.com/stephaschu/qwc2/commit/77367800f687db961f94291693022027df4e2516

Also if the request is called with "text/html", chrome does not render the result. The onload event at iframe showing the result is somewhow not not firing, so i added srcDoc: https://github.com/stephaschu/qwc2/commit/a72e681e3a8454d22e0d29e25e5a7a8e7b301142

### DIFF
--- a/utils/ServiceLayerUtils.js
+++ b/utils/ServiceLayerUtils.js
@@ -46,6 +46,10 @@ const ServiceLayerUtils = {
         } catch(e) {
             infoFormats = ['text/plain'];
         }
+        // prefer text/html
+        if(infoFormats.includes('text/html')) {
+            infoFormats = ['text/html'];
+        }
         let topLayer = null;
         let serviceUrl = null;
         try {
--- a/components/IdentifyViewer.jsx
+++ b/components/IdentifyViewer.jsx
@@ -263,7 +263,7 @@ class IdentifyViewer extends React.Component {
             );
         } else if(result.type === "html") {
             resultbox = (
-                <iframe className="identify-result-box" onLoad={ev => this.setIframeContent(ev.target, result.text)}></iframe>
+                <iframe className="identify-result-box" srcDoc={result.text} onLoad={ev => this.setIframeContent(ev.target, result.text)}></iframe>
             );
         } else if(result.properties.htmlContent) {
             if(result.properties.htmlContentInline) {


utils/ServiceLayerUtils.js
```diff
@@ -46,6 +46,10 @@ const ServiceLayerUtils = {
         } catch(e) {
             infoFormats = ['text/plain'];
         }
+        // prefer text/html
+        if(infoFormats.includes('text/html')) {
+            infoFormats = ['text/html'];
+        }
```
elemoine: I don't know yet how to modify the QWC2 code to accomodate your use case, but the above change doesn't look correct to me. Here `infoFormats` contains all the GetFeatureInfo formats supported by the WMS service, and I don't think it makes sense to limit it to `text/html` when `text/html` is one of the supported formats. This would break existing applications.

stephaschu: You're right. Also this might break if the WMS supports text/html but there is nothing configured (i.e. mapserver template).

manisandro:
Not sure I completely understand the usecase you describe, can you post an example of an external layer with which one can observe this?

stephaschu:
WMS Url: 
https://www.umweltkarten-niedersachsen.de/arcgis/services/Natur_wms/MapServer/WMSServer?
FeatureInfo: 
https://www.umweltkarten-niedersachsen.de/arcgis/services/Natur_wms/MapServer/WmsServer?service=WMS&version=1.3.0&request=GetFeatureInfo&id=Naturschutzgebiet1560167819966&layers=Naturschutzgebiet&query_layers=Naturschutzgebiet&x=51&y=51&i=51&j=51&height=101&width=101&srs=EPSG:25832&crs=EPSG:25832&bbox=367876.6666666666,5929471.249999999,373221.25,5934815.833333333&info_format=text%2Fxml&with_geometry=true&with_maptip=false&feature_count=20&FI_LINE_TOLERANCE=8&FI_POINT_TOLERANCE=16&FI_POLYGON_TOLERANCE=4

elemoine:
@manisandro the issue is that the text/xml info format is selected and that the XML returned by the external WMS is not something that QWC2 can parse/understand.
Maybe we could add a preferredInforFormat option or something at the theme item level?

stephaschu:
@elemoine you can force "text/html" by adding "info_format": "text/html" at https://github.com/qgis/qwc2-demo-app/blob/bf49843c858281fd8f314f84e900d42d17d8cae3/config.json#L342 But this makes all FeatureInfo requests called with "text/html".

elemoine:
@stephaschu yep, not ideal.

manisandro:
I suppose it could make sense to switch to text/html (if possible, else text/plain) as default for external WMS layers, which may very well not be served by QGIS Server.

stephaschu:
The QGIS Server behind qwc2-demo-app supports text/html:

http://qwc2.sourcepole.ch/wms/wolfsburg/qwc2_demo?service=WMS&version=1.3.0&request=GetFeatureInfo&id=Wolfsburg1560762115649&layers=Beschriftung&query_layers=Beschriftung&x=51&y=51&i=51&j=51&height=101&width=101&srs=EPSG:3857&crs=EPSG:3857&bbox=1196865.0946211782,6882698.524485831,1201092.6640695124,6886926.093934164&info_format=text/html&with_geometry=true&with_maptip=false&feature_count=20&FI_POINT_TOLERANCE=16&FI_LINE_TOLERANCE=8&FI_POLYGON_TOLERANCE=4

So QGIS Server as external WMS shouldn't be an issue.

manisandro:
Yes, what I meant is that QWC2 expects the text/xml feature info response in QGIS Server format, but since external layers may very well not be served by a QGIS Server, it would make sense to default to text/html or text/plain for external WMS layers.

elemoine:
> Yes, what I meant is that QWC2 expects the text/xml feature info response in QGIS Server format, but since external layers may very well not be served by a QGIS Server, it would make sense to default to text/html or text/plain for external WMS layers.

Yep, that makes sense to me. Or at least change the priorities, and have text/xml be last in the priority list.

manisandro:
Should be fixed with https://github.com/qgis/qwc2/commit/1aebf6a7da363db40234f291f4e87ddfd4b35833

PR #79, opened 2019-06-13
Use history.replaceState rather than pushState

elemoine:
This PR suggests to use history.replaceState rather than history.pushState in PermalinkUtils.js. This is to update the URL in the address bar without actually pushing a new history entry to the history stack.

There is a lot more that could be done using the history API. This PR just fixes the issue reported with https://github.com/qgis/qwc2-demo-app/issues/148.


manisandro:
Thanks!

PR #81, opened 2019-07-16
fix levelcollapse

tobias-brunner:


tobias-brunner:
after a second thought, I guess that the .py file was always working.
and the js could have been a single line:

getLayerTree(subLayer, layerEntry.sublayers, visibleLayers, printLayers, level + 1, collapseBelowLevel, titleNameMap, featureReports);

will add a new pull request.

PR #84, opened 2019-07-18
change map on click with singleclick

luciman:
I propose to change 
`map.on('click', (event) => {`
 with 
`map.on('singleclick', (event) => {` 
because it is very annoying when the DoubleClickZoom is true especially on mobile


manisandro:
Makes sense, thanks!

PR #85, opened 2019-07-20
wmts reprojection

stephaschu:
When using a custom projection like epsg:25832 as mapCrs and adding wmts layer in epsg:3857, 
the wmts layer not getting reprojected to epsg:25832. 
As stated here https://openlayers.org/en/latest/doc/tutorials/raster-reprojection.html source needs a projection with an extent for the custom projection to be reprojected.


manisandro:
Thanks!

PR #88, opened 2019-08-30
fix layergroup tristate

stephaschu:
getGroupVisibility should return a number otherwise groups end up in tristate if unchecked


manisandro:
Thanks!

PR #89, opened 2019-09-22
Add missing dependency for RedliningBufferSupport

stephaschu:
also the icon is missing in https://github.com/qgis/qwc2/tree/master/icons
could you please add it? nice feature btw


manisandro:
Thanks - yes I'll also add the icon.

PR #90, opened 2019-10-17
fix Uncaught TypeError: features.forEach is not a function when expor…

luciman:
fix 

> Uncaught TypeError: features.forEach is not a function 

when export layer or feature if exportFormat is csv
I don't know if it is the best solution but it fix it


manisandro:
Can you post a GetFeatureInfo response which triggers the issue?

manisandro:
Fixed in https://github.com/qgis/qwc2/commit/c08b819a8775c18088a22cae880511b4cf453ed6, thanks!

PR #94, opened 2020-02-13
Allow overwriting of legendUrl, featureInfoUrl and printUrl

marioba:
Allows to manually overwrite the values:
- `legendUrl`
- `featureInfoUrl`
- `printUrl`

in the `themesConfig.json` file.

This is useful if you want to use the same QGIS project on different systems, using different `themesConfig.json`files, but without changing the advertised URL in the QGIS project each time.


manisandro:
Thanks!

PR #95, opened 2020-02-18
Fix production build on Windows

lpinca:
Remove the single quotes from the value of the `NODE_ENV` environment
variable.

### DIFF
--- a/scripts/package-commands.bat
+++ b/scripts/package-commands.bat
@@ -3,7 +3,7 @@ IF "%1"=="build" (
 npm run clean && ^
 npm run iconfont && ^
 mkdir dist && ^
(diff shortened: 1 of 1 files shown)


lpinca:
```
PS C:\Users\luigi\test> cat .\test.bat
cmd /C "set NODE_ENV='production' && node ./test.js"
PS C:\Users\luigi\test> cat test.js
console.log(process.env.NODE_ENV);
console.log(process.env.NODE_ENV === 'production');
PS C:\Users\luigi\test> .\test.bat

C:\Users\luigi\test>cmd /C "set NODE_ENV='production' && node ./test.js"
'production'
false
PS C:\Users\luigi\test>
```

manisandro:
Thanks!

PR #97, opened 2020-02-28
Add XYZ tiles for Google Maps

lpinca:
Add Google Maps to `CONFIGPROVIDER`.


pka:
Thanks for your PR. But since Google forbids using Google Maps via direct tile requests, we can't merge this, sorry.

PR #100, opened 2020-04-02
set empty date value to null instead of empty string

luciman:
set empty date value to null instead of empty string even if the field is left untouched not just onChange and avoid 

> Warning: `value` prop on `input` should not be null.

>

### DIFF
--- a/components/AutoEditForm.jsx
+++ b/components/AutoEditForm.jsx
@@ -88,8 +88,7 @@ class AutoEditForm extends(React.Component) {
                 <span className="input-frame">
                     <input name={field.id} type={field.type} {...constraints}
                         onChange={(ev) => {
-                            // set empty date field value to null instead of empty string
-                            this.props.updateField(field.id, ev.target.value == '' ? null : ev.target.value);
+                            this.props.updateField(field.id, ev.target.value);
                         }}
                         value={value}/>
                 </span>
--- a/components/widgets/style/ToggleSwitch.css
+++ b/components/widgets/style/ToggleSwitch.css
@@ -47,3 +47,8 @@ div.ToggleSwitch.toggle-switch-active > span.toggle-switch-no {
 div.ToggleSwitch.toggle-switch-inactive > span.toggle-switch-yes {
     display: none;
 }
+
+
+div.ToggleSwitch > input {
+    display: none;
+}
--- a/plugins/Editing.jsx
+++ b/plugins/Editing.jsx
@@ -246,6 +246,8 @@ class Editing extends React.Component {
             let element = ev.target.elements.namedItem(name);
             if(element) {
                 let value = element.type === "radio" || element.type === "checkbox" ? element.checked : element.value;
+                // set empty date value to null instead of empty string
+                if (element.type === "date" && element.value === "") value = null;
                 if(feature.properties[name] === undefined) {
                     feature.properties[name] = value;
                 }


manisandro:
Ups, that part got lost somewhere. Thanks!

manisandro:
Actually I was meaning to merge #101, which as I see contains the second commit in this PR.

Can you explain when you get the error you are addressing with the first commit?

luciman:
@manisandro there is the case when the user let the date input untouched and the submit is with "" empty string. The empty string is replaced with null just onChange 
https://github.com/qgis/qwc2/blob/7f6269aef2ca7938866d9281dc701925484c45e4/components/AutoEditForm.jsx#L92

if the value would have been checked and initialized with null here 
https://github.com/qgis/qwc2/blob/7f6269aef2ca7938866d9281dc701925484c45e4/components/AutoEditForm.jsx#L86
or here 
https://github.com/qgis/qwc2/blob/7f6269aef2ca7938866d9281dc701925484c45e4/components/AutoEditForm.jsx#L94 
we would have received: `Warning: value prop on input should not be null.`
that's why I chose to do the check the value on submit function

manisandro:
But it should never happen that `value` is `null` there, since it is being set to an empty string in that case here:

https://github.com/qgis/qwc2/blob/7f6269aef2ca7938866d9281dc701925484c45e4/components/AutoEditForm.jsx#L37

luciman:
yes it is not happen. maybe I don't express myself well. the idea is that the date input value should never be an empty string, it should be null. I think the same logic was at @mwa 's https://github.com/qgis/qwc2/commit/098f85f9566f2363efbc035a3734a33cc41c4400  transformed in https://github.com/qgis/qwc2/blob/7f6269aef2ca7938866d9281dc701925484c45e4/components/AutoEditForm.jsx#L92
but the problem is that the empty string is replaced with null just onChange and if the user don't make any change it remain empty string and it should be null

manisandro:
OK understood thanks - but you'll still need this one https://github.com/qgis/qwc2/blob/7f6269aef2ca7938866d9281dc701925484c45e4/components/AutoEditForm.jsx#L92
because the the logic here https://github.com/qgis/qwc2/blob/16878628e6e810e2dbb325528e2096aa9c7ed729/plugins/Editing.jsx#L249 only actually set the date to `null` if the feature property is unset, so i.e. when drawing a new feature and committing without touching the date input. But if you have a date set and you clear it, you still want the line above to ensure that the value is transformed to null.

Committed as https://github.com/qgis/qwc2/commit/16878628e6e810e2dbb325528e2096aa9c7ed729

luciman:
Thanks
I don't think                           https://github.com/qgis/qwc2/blob/7f6269aef2ca7938866d9281dc701925484c45e4/components/AutoEditForm.jsx#L92 can become `this.props.updateField(field.id, ev.target.value);` because also if you have the date and you clear it ... the date value will be an empty string and on submit it will be transformed to null here                 https://github.com/qgis/qwc2/blob/16878628e6e810e2dbb325528e2096aa9c7ed729/plugins/Editing.jsx#L249

manisandro:
Note that here https://github.com/qgis/qwc2/blob/16878628e6e810e2dbb325528e2096aa9c7ed729/plugins/Editing.jsx#L253
the transformed value is only used if the property is `undefined`, which is not the case as soon as the date field is touched in any way.

tpo:
Please excuse my intrusion from the side lines (I'm "watching" the project, so I hear what you're writing).

Your whole conversation seems to crying "implicit logic" to me. Meaning there is a very strong interdependency between functional parts with no clear and **explicit** interfaces exposing those interdependencies. And even worse these implicit interdependencies do not seem to be documented _at all_!

If I may, then I'd suggest to __both__ make those interdependencies __explicit__ in the interfaces between the functional parts (other languages offer pre/post conditions for function arguments, in JS that'd probably translate to explicit argument checks) __and__ in addition to document those interdependencies in the documentation of the relevant functions.

I hope my interference her isn't mere noise and actually adds some signal. Please accept my sincer apology if it turns out the be the former :-/

PR #102, opened 2020-04-20
Freehand drawing option in Redlining tool

benoitblanc:
Add a button in Redlining tool to enable freehand drawing for LineString and Polygon geometries

Fix https://github.com/qgis/qwc2-demo-app/issues/193


manisandro:
Looks good, thanks!

PR #107, opened 2020-05-19
Make url and projection configurable for osm sources

autra:
# Problem

When using layers of `osm` type I'd like to be able to use `tile.openstreetmap.fr` instead of `.org`. 

# Solution

Add support for `options.url` and `options.projection` when creating the `OSMSource`.

Thanks


manisandro:
Thanks

PR #110, opened 2020-05-29
add constraints to all inputs

luciman:
even if in the test there are examples for constraints for ex numeric input -> readOnly or required they were not implemented
the style solution for ToggleSwitch checkbox is a little creepy but on display:none the input is not focusable and the browser can not display the message "Please fill out this field" next to it


manisandro:
Thanks!

PR #111, opened 2020-06-03
check if editing selected layer is visible/ make visibility true if i…

luciman:
…t is not / restore initial layer visibility state on change selected layer or exit editing

I think it would be useful if the layer selected for editing would become automatically visible if it is not, and when finishing editing or changing the layer for editing to return to the initial state

### DIFF
--- a/plugins/Editing.jsx
+++ b/plugins/Editing.jsx
@@ -70,23 +70,6 @@ class Editing extends React.Component {
         this.setState({selectedLayerVisibility: null});
     }
 
-    findParents(layer, title) {
-        if (layer) {
-            if (layer.title === title) {
-                return []
-            }
-            if (Array.isArray(layer.sublayers)) {
-                for (let i = 0; i < layer.sublayers.length; i++) {
-                    const childResult = this.findParents(layer.sublayers[i], title)
-                    if (Array.isArray(childResult)) {
-                        return [i].concat(childResult);
-                    }
-                }
-            }
-        }
-    }
-
-
     componentWillReceiveProps(newProps) {
         let themeSublayers = newProps.layers.reduce((accum, layer) => {
             return layer.role === LayerRole.THEME ? accum.concat(LayerUtils.getSublayerNames(layer)) : accum;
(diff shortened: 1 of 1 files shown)


plugins/Editing.jsx
```diff
@@ -233,10 +264,30 @@ class Editing extends React.Component {
             </SideBar>
         );
     }
+
+    checkVisibility = (selectedLayer, visibility) => {
+        if (selectedLayer != null){
+            let layer = this.getLayer(this.props.layers)
```
manisandro, line 270: This won't work if the theme layer is split (i.e. when reordering is active an an external layer is placed in the middle).

You'll want to do something like

    let path = null;
    this.props.layers.filter(layer => (layer.role === LayerRole.THEME && LayerUtils.searchSubLayer(layer, 'name', selectedLayer, path));

and you'll also get the path for free.

luciman:
@manisandro You are right. I changed, I hope it's OK, thank you!

plugins/Editing.jsx
```diff
@@ -233,10 +255,30 @@ class Editing extends React.Component {
             </SideBar>
         );
     }
+
+    checkVisibility = (selectedLayer, visibility) => {
+        if (selectedLayer != null){
+            let layer = this.props.layers.filter(layer => (layer.role === LayerRole.THEME && LayerUtils.searchSubLayer(layer, 'name', selectedLayer)))[0];
+            let path = this.findParents(layer, selectedLayer);
```
manisandro, line 262: Don't you want to use the path returned by LayerUtils.searchSubLayer?

manisandro: Actually you can also use this.props.layer.find instead of filter.

luciman: thank you for your guidance

manisandro:
I've made some robustness changes:
- in onHide, you cleared the selectedLayerVisibility state, but the selectedLayer itself is not cleared. So I suggest just keeping also the selectedLayerVisibility state
- `this.props.layers.filter`  ->  `this.props.layers.find`, testing that the layer is actually found

Please check whether everything still works though

luciman:
It works! Thank you

manisandro:
Thanks!

PR #115, opened 2020-06-11
Add geojson export format

FlavBd:
Hello !

Like I mentionned in https://github.com/qgis/qwc2-demo-app/issues/198  that I work on the possibility to have different export format when we select features in QWC2. 

Here, I submit the geojson format.

### DIFF
--- a/components/IdentifyViewer.jsx
+++ b/components/IdentifyViewer.jsx
@@ -230,14 +230,11 @@ class IdentifyViewer extends React.Component {
             Object.entries(json).forEach(([layerName, features]) => {
                 let data = {
                     "type":"FeatureCollection",
-                    "features":[]
+                    "features": features
                 };
-                let datageojson = {};
-                features.forEach(feature => {
-                    data.features = data.features.concat(feature);
-                    datageojson = JSON.stringify(data, null, ' ');
-                });
-                FileSaver.saveAs(new Blob([datageojson], {type: "application/geo+json;charset=utf-8"}), layerName + ".geojson");
+                FileSaver.saveAs(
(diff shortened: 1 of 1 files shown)


components/IdentifyViewer.jsx
```diff
@@ -226,6 +226,19 @@ class IdentifyViewer extends React.Component {
         if(this.state.exportFormat === 'json') {
             let data = JSON.stringify(json, null, ' ');
             FileSaver.saveAs(new Blob([data], {type: "text/plain;charset=utf-8"}), "results.json");
+        } else if(this.state.exportFormat === 'geojson') {
+            Object.entries(json).forEach(([layerName, features]) => {
+                let data = {
+                    "type":"FeatureCollection",
+                    "features":[]
```
manisandro, line 233: Why not just

    "features": features

?

FlavBd: I tried to put` "features": feature` but this gives me `[object Object]` in my geojson files even if I put `JSON.stringify` after `let data`. Also I need to look over all the features to not have the layer's name in the geojson.

manisandro: You are already separating features from layernames by iterating over `Object.entries(json)`. This

    features.forEach(feature => {
                    data.features = data.features.concat(feature);
    })

is really the same as `data.features = features`. And you also shouldn't need to reserialize the data in that loop.

manisandro:
Thanks

components/IdentifyViewer.jsx
FlavBd: Sorry, I didn't understand at the first comment but now I changed the code. Thanks for your advices.

PR #117, opened 2020-08-19
Inform user where there are in production script

stev-0:
It's currently confusing when running a command like `yarn run prod` as right before the webpack command (which takes some time) there is the statement `Done!`. To make it less confusing for people perhaps a line could be added to explain what is going on, as webpack doesn't seem to provide any info (or alternatively you could provide more verbose output from webpack).


manisandro:
Make sense, thanks!

PR #123, opened 2020-10-23
fix identifyViewer attribValue

luciman:
after https://github.com/qgis/qwc2/commit/5799b5020ba2b66bafd751bd33db761c8cda3637 the Attribute value is the same with Attribute name


manisandro:
Thanks!

PR #124, opened 2021-01-20
downgrade react-rnd so that touch event works on mobile devices

tudorbarascu:
See https://github.com/bokuweb/react-rnd/issues/760


manisandro:
Thanks!

PR #125, opened 2021-01-21
Pass the correct sublayer to maxbbox()

lpinca:
This fixes a bug introduced in da23d28ab6. `sublayer` is no longer a key
but a reference to an element in the `layer.sublayers` array.


manisandro:
Thanks!

PR #126, opened 2021-01-21
use ol.proj for reprojection and get units

luciman:
for fixing https://github.com/qgis/qwc2-demo-app/issues/254 and also for consistency I suggest using ol.proj.transform instead of Proj4js.transform.
from https://github.com/proj4js/proj4js/commit/e4e030847c70a50a81db8704818c58ff6ba96cfa  there is a problem with the transformation on neu axis

proj4: "2.6.0" or less,
Proj4js.transform
EPSG:4326 {x: 23.569485, y: 44.483493}   ---->  EPSG:3844 {x: 386331.41681077797 y: 332533.2419243645}
  
proj4@2.6.1 or higher
 Proj4js.transform
EPSG:4326 {x: 23.569485, y: 44.483493}  ---->  EPSG:3844 {y: 386331.41681077797, x: 332533.2419243645}

there also was a problem with this on WMTS https://github.com/openlayers/openlayers/issues/11076 but it seems it is ok with WMTS rendering

Thanks

### DIFF
--- a/components/CoordinateDisplayer.jsx
+++ b/components/CoordinateDisplayer.jsx
@@ -9,7 +9,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
-import proj4js from 'proj4';
 import CoordinatesUtils from '../utils/CoordinatesUtils';
 import LocaleUtils from '../utils/LocaleUtils';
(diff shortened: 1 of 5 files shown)


manisandro:
Thanks!

PR #128, opened 2021-01-29
update @turf/buffer dependency

stephaschu:
There was an issue with calculating distance in turfjs buffer function (i.e. https://github.com/Turfjs/turf/issues/1246)
With the updated version the distance seems correct.


manisandro:
Thanks!

PR #129, opened 2021-02-03
Fix deactivating Toolbar item

stephaschu:
This fixes an issue when a Toolbar item could not be deactivated if it has no mode specified.

item.mode is "undefined" while this.props.currentTaskMode is "null" which made it never become active.


manisandro:
Thanks!

PR #130, opened 2021-03-04
Add optional parameter FORMAT_OPTIONS in DxfExport

stephaschu:
i.e. https://docs.qgis.org/3.16/en/docs/server_manual/services.html#format-options


manisandro:
Thanks

PR #131, opened 2021-04-21
Manage bookmarks

benoitblanc:
Hi @manisandro 

I have just finished the work to manage bookmarks as discussed in https://github.com/qwc-services/qwc-permalink-service/pull/5

To do so, I have created a new menu item that sould be available only for logged-in users. That is why I created a new `requireAuth` option in menu items for the configuration. Then, I created a plugin and a window to list bookmarks for currently logged-in user. Here are some screenshots :

* Bookmarks are available only for logged-in users : 
image image

* Bookmarks are listed in the window. User can add a new one with a description using the current map :
image 
image

* User can open bookmarks in new tab, overwrite bookmark with the current map, or remove bookmark.
image

This plugin uses the work done in https://github.com/qwc-services/qwc-permalink-service/pull/5 and https://github.com/qwc-services/qwc-config-db/pull/2 to create, get, remove bookmarks in QWC configuration database.

Please let me know if you have some remarks on what I have done.

Thank you for your review and your feedback


manisandro:
Thanks, looks good at a first glance! Slight nit pick: would it be possible to style the panel a bit more consistently with the other panels? I.e. taking the Share panel as a template for the heading style and padding etc.

benoitblanc:
I have changed heading with `h4` and added padding for `div` in the window.

image

components/BookmarkWindow.jsx
```diff
@@ -0,0 +1,102 @@
+/**
+ * Copyright 2016-2021 Sourcepole AG
```
manisandro, line 2: You'll probably want to change the Copyright?

plugins/Bookmark.jsx
```diff
@@ -0,0 +1,37 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import {connect} from 'react-redux';
+import PropTypes from 'prop-types';
+import ConfigUtils from '../utils/ConfigUtils';
+import BookmarkWindow from '../components/BookmarkWindow';
+ 
+class Bookmark extends React.Component {
```
manisandro, line 15: Is there a reason for creating the wrapping plugin for the component instead of just having the entire component as the plugin in one single file? Also, what do you think about using a sidebar (like the Share plugin) rather than a window?

benoitblanc: @manisandro I think I have created the plugin to be able to enable/disable it in `config.json` file. But I realise it may be useless and I should have the component as the plugin ?

I tried to use a sidebar and a window for my tests. I just preferred the window because it is resizeable and dockable but it can be discussed :)

manisandro: Yes you can spare yourself the boilerplate and just expose the component itself as a plugin.

Personally I'd go with a sidebar for consistency with the other functionalities, but your call.

benoitblanc: @manisandro I did the changes to use component directly as a plugin.

Also, I think you were right about the SideBar, it is more consistent for the user to get all the functionalities at the same place :+1:

manisandro:
Thanks!

PR #132, opened 2021-04-27
Add XYZ layer to set up its own provider for background layers

benoitblanc:
Hi @manisandro ,

I want to be able to add my own provider for background layers, without add it in `utils/ConfigProvider.js`. To do so, I need to add XYZLayer with `ol.source.XYZ` configured with the right URL.

Maybe I will need to add some additional options to the source later.

Thanks


manisandro:
Thanks

PR #133, opened 2021-04-28
Remove node-fqdn dependency

lpinca:
The reason is that `node-fqdn` depends on `deasync` which needs to be
compiled if a prebuilt binary is not available. Doing this is not always
possible/easy, especially on Windows.


manisandro:
Thanks

PR #136, opened 2021-05-19
restore non fractional zoom to it's previous behavior

stephaschu:
With "allowFractionalZoom": false in config.json zooming in/out was kind of clunky. In example when zooming in with mousewheel from 1:10000, the scale was still at 1:10000 but the map was zooming in a bit, so zooming to next fixed scale needed a few mousewheel pushes.

By adding the constrainResolution to MouseWheelZoom interaction the previous behavior to get to the next fixed scale should be restored.


manisandro:
Thanks

PR #139, opened 2021-07-12
Fix printing of additional theme layers

mwa:
Additional theme layers could not be printed as external layers, as they were using only relative WMS URLs and incorrect WMS sublayers.

Fixes:
* change relative URLs of theme layer to full URL on current location
* ~~get theme sublayers from `layer.params.LAYERS` instead of `layer.name`~~ (cf. 007437fc)
* ~~add a separate external print layer for each theme sublayer to support individual layer opacities~~ (cf. 007437fc)

Limitations:
* the full URL must be reachable from QGIS Server to render an external layer
  * Note: use `http://172.17.0.1:8088/` for testing with `qwc-docker` on `localhost`
* only public theme layers can be printed, as the identity is not forwarded


manisandro (approved):

manisandro:
Manually merged

PR #142, opened 2021-08-05
add support for show layertree on startup

mittamies:
I added support for show layertree on startup.

New config.json option below appMenuVisibleOnStartup: 
"layerTreeVisibleOnStartup": true,


Farkadi06 (approved):

manisandro:
Sorry for the late reply - maybe a better idea would be to have a `startupTask` setting on config.json, which is then activated via `setCurrentTask` in the `AppInitComponent`. It would be more flexible, as this approach here would conflict if i.e. another component gets a "...VisibleOnStartup" config prop.

PR #144, opened 2021-08-24
Update translations for germany

stamogis:
Update german translations for germany. According to the german orthography in germany for example "Maßstab" with "ß" (Eszett).


manisandro:
Thanks

PR #145, opened 2021-08-24
Update tooltips

stamogis:
Add additional tooltips and optimize existing ones.


components/Search.jsx
```diff
@@ -207,7 +212,7 @@ class Search extends React.Component {
                         'searchbar-provider-selection-active': isEmpty(this.props.activeProviders)
                     });
                     allEntry = (
-                        <li className={itemClass} key="all" onClick={() => this.props.changeSearch("", null)}>{LocaleUtils.tr("search.all")}</li>
+                        <li className={itemClass} key="all" onClick={() => this.props.changeSearch("", null)} title="">{LocaleUtils.tr("search.all")}</li>
```
manisandro, line 215: What's the purpose of the empty `title` here and elsewhere?

stamogis: This is to hide the tooltip of the button in the selection list of search providers. An alternative could be the name of the individual search provider.

manisandro:
Thanks

PR #146, opened 2021-08-24
Skip htmlContentInline in csv export

stamogis:
Skip `htmlContentInline` similar to `htmlContent` in csv export.


manisandro:
Thanks

PR #147, opened 2021-08-24
Fix invisible layer query for layer groups

stamogis:
Also activate the parent layer group with the "enable invisible layer" function of the search.


manisandro:
Thanks

PR #151, opened 2021-09-28
allow set position of sidebar

vvmruder:
I offer this to the main project. Through this, it will be possible to configure if sidebar will be on the left or the right side of the application. This is necessary because plain CSS won't preserve correct transition of menu collapsing and expanding since this values are produced dynamically.

Right side stays default.

Configuration option is available through all "sidebar" elements with the prop `side` and one of the values `"right"`/`"left"`.
image


manisandro:
Thanks

PR #153, opened 2021-10-01
add startup task

vvmruder:
Yet another offering for the main project. Since I had the same requirement for the viewer like it was asked here: https://github.com/qgis/qwc2/pull/142

I tried to solve it like suggested. Hopefully I did it right. Feedback is very welcome.

As suggested in config.json a new setting is added `"startupTask"`

I tested it with different tasks like `{"key": "LayerTree"}` and `{"key": "Measure", "mode": "Polygon", "mapClickAction": "identify"}`

Especially the last one let me think, that I did not solved it completely right because the identify plugin is always triggered while measuring. I couldn't stop it from doing that.

Cheers
C

### DIFF
--- a/components/StandardApp.jsx
+++ b/components/StandardApp.jsx
@@ -141,8 +141,7 @@ class AppInitComponent extends React.Component {
                 }
             });
         });
-        const task = ConfigUtils.getConfigProp("startupTask")
-        console.log(task);
+        const task = ConfigUtils.getConfigProp("startupTask");
         if(task){
             this.props.setCurrentTask(task.key, task.mode, task.mapClickAction);
         }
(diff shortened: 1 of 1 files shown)


vvmruder:
May I ask if this has to be improved?

components/StandardApp.jsx
```diff
@@ -140,6 +141,12 @@ class AppInitComponent extends React.Component {
                 }
             });
         });
+        const task = ConfigUtils.getConfigProp("startupTask")
+        console.log(task);
```
manisandro, line 145: You'll want to drop this line

vvmruder: well ok :)  => Done 

sorry for this left over.

vvmruder:
Is it now ok from your perspective?

manisandro:
Yes - thanks!

PR #154, opened 2021-10-07
add ThemeUtils.fullUrl to externalLayer Url

stephaschu:
fixes an issue when printing backgroundLayer as EXTERNAL_WMS with relative Urls


manisandro:
Thanks

PR #155, opened 2021-10-13
fix identify features in user vector layers

luciman:
identify features on user vector layers not working. see gifs

Before this pull it can be seen that identify is not working as expected.
identify before
:

After
identify after


manisandro:
Thanks

PR #156, opened 2021-10-13
fix get WMTS Layer if Dimension or ServiceProvider is missing

luciman:
there may be wmts layers without Dimension like 
https://basemap.at/wmts/1.0.0/WMTSCapabilities.xml

and also Capabilities without ServiceProvider


manisandro:
Thanks

PR #167, opened 2021-11-03
keep the same zIndex also when ResizeableWindow is maximized

luciman:
otherwise when you'll open an IframeDialog from IdentifyViewer and you will mazimized it, the IframeDialog goes backwards IdentifyViewer


manisandro:
Thanks

PR #170, opened 2021-11-26
Fix plugin name in translations

lpinca:
Refs: https://github.com/qgis/qwc2/commit/e08aed5865534044
Refs: https://github.com/qgis/qwc2-demo-app/commit/f8e29a5126fa90d0

### DIFF
--- a/translations/cs-CZ.json
+++ b/translations/cs-CZ.json
@@ -9,7 +9,7 @@
         "DxfExport": "DXF Export",
         "Editing": "Úpravy",
         "Help": "Nápověda",
-        "IdentifyRegion": "Určit oblast",
+        "Identify": "Určit oblast",
         "LayerCatalog": "",
         "LayerTree": "Vrstvy & Legenda",
         "Login": "Přihlášení",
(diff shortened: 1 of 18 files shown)


manisandro:
`IdentifyRegion` is the correct translation key, as the tool mode is appended to the plugin name [1].

[1] https://github.com/qgis/qwc2/blob/master/components/AppMenu.jsx#L120

lpinca:
This https://github.com/qgis/qwc2/blob/cc063c9f2f36e7c35860dfe4d84f7c813dfb5c83/components/Toolbar.jsx#L31 is currently not properly translated in `qwc2-demo-app`.

manisandro:
Uh right - it should build the translation key the same way AppMenu does

manisandro:
Fixed in https://github.com/qgis/qwc2/commit/47bfd1017cdc2aad659fb07c19c71d8af206bba0, thanks for reporting!

lpinca:
@manisandro can you please take a look at https://github.com/qgis/qwc2-demo-app/pull/299?

PR #172, opened 2021-12-03
fix AttributeTable numeric input onChange

luciman:
on input type number nr is not the value. I think there was a confusion with NumericInput from  react-numeric-input2 which is used in AutoEditForm


manisandro:
thanks!

PR #173, opened 2021-12-14
add catalan translations

edgarfuste:
Improvement: 

Add translation file "ca-ES.json" for Catalan language into translations folder.


manisandro:
Thanks, you'll also want to add it to the list of languages here https://github.com/qgis/qwc2/blob/master/translations/tsconfig.json

manisandro:
Thanks!

PR #174, opened 2021-12-28
fix AttributeTable first column style

luciman:
Screenshot from 2021-12-28 12-34-32

fix first column sticky position


manisandro:
Thanks!

PR #177, opened 2022-01-20
fix Redlining/RedliningBufferSupport VectorLayerPicker select label s…

luciman:
…tyle

Screenshot from 2022-01-20 16-37-00


manisandro:
Thanks

PR #178, opened 2022-01-28
use the same openExternalUrl function for Toolbar items like in AppMenu

luciman:
posibility to open showIframeDialog  for Toolbar items like in AppMenu if you have
....
 {"key": "ExternalLink", "icon": "link", "target": "iframe", "url": "http://example.com?extent=$e$"}
.....


manisandro:
Thanks!

PR #179, opened 2022-01-31
German and Swiss translation updated in terms of grammar and wording

danceb:
I have corrected some spelling mistakes and changed a few wordings for better readability. In addition the locale of the Swiss translation was named as "de-DE" and not "de-CH".


manisandro:
Thanks!

PR #180, opened 2022-02-01
Fix URL joining

lassitanskanen:
Fixing URL joining if use parameters with original url.

Original themesConfig.json url: https://domain.se/qgis-server?MAP=/data/test.qgs

Before:
https://domain.se/qgis-server?MAP=/data/test.qgs?SERVICE=WMS&REQUEST=GetProjectSettings
After:
https://domain.se/qgis-server?MAP=/data/test.qgs&SERVICE=WMS&REQUEST=GetProjectSettings

### DIFF
--- a/scripts/themesConfig.py
+++ b/scripts/themesConfig.py
@@ -12,7 +12,7 @@ try:
 except:
     import urllib2 as request
 try:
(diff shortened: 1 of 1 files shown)


manisandro:
Thanks

PR #181, opened 2022-02-08
fix typo from update_params and modify urljoin in getThumbnail also

lassitanskanen:
Typo in _replace method and use better params joining in getThumbnail also.


manisandro:
Thanks

PR #183, opened 2022-02-17
Make search optional in TopBar plugin

vvmruder:
This change checks if Search element was provided. If not, TopBar does not contain a search.

Only down side is: All elements right next to the search are sliding to the most left next to the logo when search does not exists. What would be a good way to avoid that?


vvmruder:
Like this one: https://github.com/qgis/qwc2/blob/c7e785988a37013f23f1dcd01b4765a4548f8cc3/plugins/style/LayerTree.css#L115-L118

PR #184, opened 2022-02-24
fix attachment fileUrl

luciman:
there should not be added / because it already exists in editServiceUrl declaration
end in all other places like    
https://github.com/qgis/qwc2/blob/00f260cb1c1d79abe88596085471684a0d01697b/utils/EditingInterface.js#L50
https://github.com/qgis/qwc2/blob/00f260cb1c1d79abe88596085471684a0d01697b/utils/EditingInterface.js#L150
 https://github.com/qgis/qwc2/blob/00f260cb1c1d79abe88596085471684a0d01697b/utils/EditingInterface.js#L139
https://github.com/qgis/qwc2/blob/00f260cb1c1d79abe88596085471684a0d01697b/utils/EditingInterface.js#L131
  is no / added


manisandro:
Thanks!

PR #186, opened 2022-02-25
Fix typo in translations/it-IT.json

lpinca:
As titled.

### DIFF
--- a/translations/it-IT.json
+++ b/translations/it-IT.json
@@ -2,7 +2,7 @@
   "locale": "it-IT",
   "messages": {
     "appmenu": {
-      "filter": "Filtra menü...",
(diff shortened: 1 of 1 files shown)


lpinca:
There is also another issue similar to https://github.com/qgis/qwc2/pull/170.

https://github.com/qgis/qwc2/blob/d249f429e949cc96fb5a7797791a45ec29fb5a91/components/Toolbar.jsx#L31

There is no translation for `appmenu.items.MeasureLineString` and `appmenu.items.MeasurePolygon` so the `title` attribute is not properly translated. I wonder if they should be added.

manisandro:
Thanks - yes would make sense to add them

PR #188, opened 2022-03-09
Add hectare unit to area measure tool

gwenandres:
Hello, 

I propose a PR to add hectare unit support to area measure tool.

Have a good day


manisandro:
Thanks

PR #189, opened 2022-03-10
Add external link in search provider results

benoitblanc:
Hi,

I would like to propose this PR to add an icon to open an external link on search provider results. I don't know if it is relevant for everyone, but you could open a link of your search provider in a new tab (for example, a link to a metadata resource).

Thanks !


manisandro:
Thanks! Might be nice to also allow for inline dialog, i.e. like here https://github.com/qgis/qwc2/blob/master/plugins/TopBar.jsx#L109 ?

benoitblanc:
@manisandro yes you are right, it could be nice to have the link opened in inline frame. I have added this possibility.
Thanks !

manisandro:
Thanks!

PR #193, opened 2022-03-31
Convert keys to uppercase

lpinca:
Before 75e63ff605 `LayerUtils.buildWMSLayerParams()` was converting
search parameters names to uppercase. In 541ae5a4ab, 75e63ff605 was
partially reverted but conversion of parameters names to uppercase was
not restored. However, other parts of the codebase depend on that
behavior. For example,
https://github.com/qgis/qwc2/blob/9f97cdc76d/utils/IdentifyUtils.js#L48.

This commit restores the original behavior.

Refs: https://github.com/qgis/qwc2/commit/75e63ff605#r69813598

### DIFF
--- a/utils/LayerUtils.js
+++ b/utils/LayerUtils.js
@@ -144,7 +144,7 @@ const LayerUtils = {
         const urlQuery = url.parse(layer.url, true).query;
         const urlParams = Object.keys(urlQuery).filter(key => {
             return !["service", "version", "request"].includes(key.toLowerCase());
-        }).reduce((res, key) => ({...res, [key]: urlQuery[key]}), {});
+        }).reduce((res, key) => ({...res, [key.toUpperCase()]: urlQuery[key]}), {});
 
         if (!Array.isArray(layer.sublayers)) {
             return {


lpinca:
We could do something like this

```diff
diff --git a/utils/IdentifyUtils.js b/utils/IdentifyUtils.js
index b638f362..c4bed211 100644
--- a/utils/IdentifyUtils.js
+++ b/utils/IdentifyUtils.js
@@ -45,7 +45,7 @@ function identifyRequestParams(layer, queryLayers, projection, params) {
             info_format: format,
             with_geometry: true,
             with_maptip: false,
-            map: layer.params.MAP,
+            map: layer.params.MAP || layer.params.map,
             ...params
         }
     };
```

but would that be sufficient? I don't know if that works for other cases where `layer.params.MAP` is used (`plugins/Map.jsx`, `plugins/Print.jsx`, and `utils/LayerUtils.js`).

manisandro:
I'll look into it

lpinca:
Another option would be to normalize only the `map` parameter

```diff
diff --git a/utils/LayerUtils.js b/utils/LayerUtils.js
index 7b8f707f..f50efe1b 100644
--- a/utils/LayerUtils.js
+++ b/utils/LayerUtils.js
@@ -144,7 +144,15 @@ const LayerUtils = {
         const urlQuery = url.parse(layer.url, true).query;
         const urlParams = Object.keys(urlQuery).filter(key => {
             return !["service", "version", "request"].includes(key.toLowerCase());
-        }).reduce((res, key) => ({...res, [key]: urlQuery[key]}), {});
+        }).reduce((res, key) => {
+            if (key === 'MAP' || key === 'map') {
+                res.MAP = urlQuery[key];
+            } else {
+                res[key] = urlQuery[key];
+            }
+
+            return res;
+        }, {});
 
         if (!Array.isArray(layer.sublayers)) {
             return {
```

but I'm not sure if it is any better.

manisandro:
https://github.com/qgis/qwc2/commit/b3b3c7cd43e41068ad944e09f16f0b077ef32247 should clean things up, and remove explicit handling of the MAP parameter, rather handling it implicitly along with any other extra parameters that may appear in the capability URLs advertised by the service capabilities (or manually specified in the called service URL).

lpinca:
Ok, I'll check it out. Thank you.

PR #194, opened 2022-04-13
Also handles changes to the transparency slider by arrow keys or mouse wheel

stamogis:
The transparency slider in the layer tree can also be moved using the arrow keys on the keyboard or with the mouse wheel (Firefox). In these cases, the layer transparency in the map was not changed. Therefore added corresponding function call for the events `onKeyUp` and `onWheel`.


manisandro:
Thanks!

PR #195, opened 2022-04-14
allow use legendUrl in external wmts layers

lassitanskanen:
Related to #339

External wmts legends will be fixed with this commit.


manisandro:
Would this also work? https://github.com/qgis/qwc2/commit/70af83df4db28edbd1f171b185f535d651a7273e

lassitanskanen:
Thanks, your commit is better.

PR #196, opened 2022-04-21
Fix for scrolling the "Map & Tools" menu with the iOS Safari browser

stamogis:
Fix for scrolling the "Map & Tools" menu with the iOS Safari browser to also reach the last entries of a longer menu when scrolling on small touch devices.


manisandro:
Thanks

PR #197, opened 2022-04-28
fix map staying blank if no permissions on default theme

olivierdalang:
This works around the issue of map staying blank if the logged user has no permission on the default map. Such a situation easily happens with the stock qwc-map-viewer and qwc-config-generator, which assumes the first project is the default project, not taking into account user permissions.

See https://github.com/qwc-services/qwc-map-viewer/issues/12 for more context.

### DIFF
--- a/components/StandardApp.jsx
+++ b/components/StandardApp.jsx
@@ -94,6 +94,20 @@ class AppInitComponent extends React.Component {
                     if (ConfigUtils.getConfigProp("dontLoadDefaultTheme")) {
                         return;
                     }
+
+                    // Determine default theme, and fallback if not available
+                    let availableThemes = themes.items.map((t) => t.id);
+                    let defaultTheme = themes.defaultTheme;
+                    if (!availableThemes.includes(themes.defaultTheme)) {
+                        console.log(`Default theme (${themes.defaultTheme}) is not available.`);
+                        if (availableThemes.length > 0) {
+                          defaultTheme = themes.items[0].id;
+                          console.log(`Falling back to ${defaultTheme}.`);
+                        } else {
+                          console.log(`No available theme to fallback to.`);
+                        }
+                    }
+
                     theme = ThemeUtils.getThemeById(themes, themes.defaultTheme);
                 }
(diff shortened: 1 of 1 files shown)


components/StandardApp.jsx
```diff
@@ -94,6 +94,20 @@ class AppInitComponent extends React.Component {
                     if (ConfigUtils.getConfigProp("dontLoadDefaultTheme")) {
                         return;
                     }
+
+                    // Determine default theme, and fallback if not available
+                    let availableThemes = themes.items.map((t) => t.id);
```
manisandro, line 99: Theme items might potentially be nested.

manisandro, line 99: Theme items might potentially be nested.

manisandro:
I wonder whether a better approach would be to be able to set the default theme per user. This way it might be hard to control which theme is shown in the end.

manisandro:
I wonder whether a better approach would be to be able to set the default theme per user. This way it might be hard to control which theme is shown in the end.

olivierdalang:
Currently (if I get things right), `defaultTheme` is populated by the config generator by just using the first theme that is found. So the proposed PR more or less matches that logic, but per user.

I agree it would be better to be able to define it per user, but where would that happen exactly ? I'm not yet too familiar with the QWC2 stack, but guess it needs to be handled by the auth backend somehow, and probably not really is attached to the user, but rather to roles. And then what happens if a user has many roles ?

IMO that's quite a big change for such a small fix, and may become quite difficult to correctly setup. By using the first one, it's deterministic so one can still use alphabetic ordering to determine which one gets loaded.

By the way, as said in https://github.com/qwc-services/qwc-map-viewer/issues/12, qwc-map-viewer could fix that on the fly instead of fixing it on the JS side.

components/StandardApp.jsx
olivierdalang: Ok will try to come up with a commit that handles nested themes. Would you have a sample themes.json test with, or some info in how to create projects that have nested themes ? Is it with QGIS themes in the project ?

(In my themes.json, I see sublayers, but nothing looking like nested themes, not even an empty list)

PR #199, opened 2022-05-20
Adaptative edit buttons

gwenandres:
Hello,

I am Gwendoline Andres, I work in Oslandia company. 

I propose this PR to deal with Edit buttons.
I work with qwc_services so I can specify if a writable layer is creatable, updatable or deletable.
But in QWC2 interface, Edit buttons (`Draw`, `Pick` and `Delete`) are always shown even if you are not allowed to make the action. 
You have a message "Dataset is not creatable" (for example) only at the end of your action. 

So I propose this PR to be able to set creatable, updatable and deletable permissions in Edit configuration file and show edit buttons or not. 

If creatable permission is set, `Draw` button in edit interface, and `Add` button in attribute table will appear. 
If updatable permission is set, `Pick` button in edit interface will be shown. 
If deletable permission is set, `Delete` buttons in edit interface and attribute table will appear, and `Pick` button in edit interface (to be abble to get `Delete` button).

I am wondering how to deal with updatable permision and edit form. I hesitate to hide `Accept` button when you are not allowed to update layer. But I think it could be confusing for users. Get the message "Dataset is not updatable" seems clearer. What do you think about that ? 

Next I will propose PR related to this one in qwc2-demo-app and qwc-config-generator.

I hope I manage to explain everything. 

Have a good day, 

Gwendoline Andres

### DIFF
--- a/plugins/AttributeTable.jsx
+++ b/plugins/AttributeTable.jsx
@@ -237,10 +237,11 @@ class AttributeTable extends React.Component {
                         <button className="button" disabled={editing || nolayer} onClick={() => this.reload()} title={LocaleUtils.tr("attribtable.reload")}>
                             <Icon icon="refresh" />
                         </button>
-                        {showAddButton ? (<button className="button" disabled={nolayer || editing || loading || layerChanged} onClick={this.addFeature} title={LocaleUtils.tr("attribtable.addfeature")}>
-                            <Icon icon="plus" />
-                        </button>
-												) : null}
+                        {showAddButton ? (
+                            <button className="button" disabled={nolayer || editing || loading || layerChanged} onClick={this.addFeature} title={LocaleUtils.tr("attribtable.addfeature")}>
+                                <Icon icon="plus" />
+                            </button>
+                        ) : null}
                         <button className="button" disabled={layerChanged || !Object.values(this.state.selectedFeatures).find(entry => entry === true)} onClick={this.zoomToSelection} title={LocaleUtils.tr("attribtable.zoomtoselection")}>
                             <Icon icon="search" />
                         </button>
@@ -258,7 +259,7 @@ class AttributeTable extends React.Component {
                             showDelButton ? (<button className="button" disabled={layerChanged || editing || !Object.values(this.state.selectedFeatures).find(entry => entry === true)} onClick={() => this.setState({confirmDelete: true})} title={LocaleUtils.tr("attribtable.deletefeatures")}>
                                 <Icon icon="trash" />
                             </button>
-														) : null
+                        ) : null
                         )}
                         {this.state.confirmDelete ? (
                             <button className="button edit-discard" onClick={() => this.setState({confirmDelete: false})}>
--- a/plugins/AttributeTable.jsx
+++ b/plugins/AttributeTable.jsx
@@ -215,8 +215,9 @@ class AttributeTable extends React.Component {
         const loading = this.state.loading;
         const editing = this.state.changedFeatureIdx !== null;
         const layerChanged = this.state.selectedLayer !== this.state.loadedLayer;
-				const showAddButton = typeof(editConfig[this.state.loadedLayer]) === 'undefined' || typeof(editConfig[this.state.loadedLayer].permissions) === 'undefined' || editConfig[this.state.loadedLayer].permissions.creatable === true;
-				const showDelButton = typeof(editConfig[this.state.loadedLayer]) === 'undefined' || typeof(editConfig[this.state.loadedLayer].permissions) === 'undefined' || editConfig[this.state.loadedLayer].permissions.deletable === true;
+        const editPermissions = (editConfig[this.state.loadedLayer] || {}).permissions || {};
+        const showAddButton = editPermissions.creatable !== false;
+        const showDelButton = editPermissions.deletable !== false;
         return (
             <ResizeableWindow dockable="bottom" icon="editing" initialHeight={480} initialWidth={800} onClose={this.onClose} title={LocaleUtils.tr("attribtable.title")}>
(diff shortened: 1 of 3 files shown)


plugins/Editing.jsx
```diff
@@ -272,7 +277,8 @@ class Editing extends React.Component {
         }
         let deleteBar = null;
         if (this.props.editing.action === 'Pick' && this.props.editing.feature && !this.props.editing.changed) {
-            if (!this.state.deleteClicked) {
+            if (!this.state.deleteClicked && (typeof(curConfig.permissions) === 'undefined' || curConfig.permissions.deletable === true)) {
+                // Delete button will appear by default if no permissions are defined in theme editConfig or when deletable permission is set
```
manisandro, line 281: This looks wrong, as I read it it will always display the YES/NO buttons rather than hiding the deleteBar completely. Also, easier:

     (curConfig.permissions || {}).deletable === true

manisandro: Or you could just do editPermissions = curConfig.permissions || {} after curConfig = ...;

gwenandres: Yes you're totally right. I will hide the deleteBar completely.

I will use `editPermissions`, thanks !
 
I added 
` typeof(curConfig.permissions) === 'undefined' `
in my if statements in order to stay compatible with previous edit config file versions without `permissions`defined. 

If no `permission` is set in edit configuration file, I want that all edit buttons are displayed. 
Do I stay in that way ?

manisandro: The fallback makes sense, but it's shorter to just write 

    const editPermissions = curConfig.permissions || {}

once at the beginning, and then testing 

    editPermissions.delteable/creatable/etc === true

afterwards. No need for typeof.

gwenandres: yes I understand :) 
Thanks for your advice

plugins/AttributeTable.jsx
```diff
@@ -215,6 +215,8 @@ class AttributeTable extends React.Component {
         const loading = this.state.loading;
         const editing = this.state.changedFeatureIdx !== null;
         const layerChanged = this.state.selectedLayer !== this.state.loadedLayer;
+				const showAddButton = typeof(editConfig[this.state.loadedLayer]) === 'undefined' || typeof(editConfig[this.state.loadedLayer].permissions) === 'undefined' || editConfig[this.state.loadedLayer].permissions.creatable === true;
+				const showDelButton = typeof(editConfig[this.state.loadedLayer]) === 'undefined' || typeof(editConfig[this.state.loadedLayer].permissions) === 'undefined' || editConfig[this.state.loadedLayer].permissions.deletable === true;
```
manisandro, line 219: Same as in Editing.jsx with editPermissions.

plugins/Editing.jsx
```diff
@@ -176,10 +177,15 @@ class Editing extends React.Component {
...
-        ];
+        const actionButtons = [];
+        if (isEmpty(editPermissions) || editPermissions.updatable === true || editPermissions.deletable === true) {
+            // Pick button will appear by default if no permissions are defined in theme editConfig or when updatable or deletable permissions are set
+            actionButtons.push({key: 'Pick', icon: 'pick', label: LocaleUtils.trmsg("editing.pick"), data: {action: 'Pick', geomReadOnly: false}});
+        }
+        if (isEmpty(editPermissions) || editPermissions.creatable === true) {
+            // Draw button will appear by default if no permissions are defined in theme editConfig or when creatable permission is set
+            actionButtons.push({key: 'Draw', icon: 'editdraw', label: LocaleUtils.trmsg("editing.draw"), data: {action: 'Draw', feature: null, geomReadOnly: false}});
+        }
```
manisandro, line 188: `editPermissions.creatable !== false` should achieve the same

gwenandres: Thank you very much for your review !

plugins/AttributeTable.jsx
```diff
@@ -234,9 +237,10 @@ class AttributeTable extends React.Component {
                         <button className="button" disabled={editing || nolayer} onClick={() => this.reload()} title={LocaleUtils.tr("attribtable.reload")}>
                             <Icon icon="refresh" />
                         </button>
-                        <button className="button" disabled={nolayer || editing || loading || layerChanged} onClick={this.addFeature} title={LocaleUtils.tr("attribtable.addfeature")}>
+                        {showAddButton ? (<button className="button" disabled={nolayer || editing || loading || layerChanged} onClick={this.addFeature} title={LocaleUtils.tr("attribtable.addfeature")}>
                             <Icon icon="plus" />
                         </button>
+												) : null}
```
manisandro, line 243: Please fix the indentation

plugins/AttributeTable.jsx
```diff
@@ -251,9 +255,10 @@ class AttributeTable extends React.Component {
                                 <span>{LocaleUtils.tr("attribtable.delete")}</span>
                             </button>
                         ) : (
-                            <button className="button" disabled={layerChanged || editing || !Object.values(this.state.selectedFeatures).find(entry => entry === true)} onClick={() => this.setState({confirmDelete: true})} title={LocaleUtils.tr("attribtable.deletefeatures")}>
+                            showDelButton ? (<button className="button" disabled={layerChanged || editing || !Object.values(this.state.selectedFeatures).find(entry => entry === true)} onClick={() => this.setState({confirmDelete: true})} title={LocaleUtils.tr("attribtable.deletefeatures")}>
                                 <Icon icon="trash" />
                             </button>
+														) : null
```
manisandro, line 261: Here also indentation - rest looks good

manisandro:
Thanks!

PR #200, opened 2022-05-23
Fix "csv + zip" export

stamogis:
Fix for broken "csv + zip" export. Change `stringify` function to `VectorLayerUtils.geoJSONGeomToWkt` for "csv + zip" export like normal csv export.


manisandro:
Thanks!

PR #203, opened 2022-05-24
Update ShareSocials.jsx

claudiadragoste:
I propose to add EmailShareButton. For the future: customizing the social networks that appear.

### DIFF
--- a/components/share/ShareSocials.jsx
+++ b/components/share/ShareSocials.jsx
@@ -14,11 +14,13 @@ import {
     LinkedinShareButton,
     TwitterShareButton,
     WhatsappShareButton,
+    EmailShareButton,
     FacebookShareCount,
     FacebookIcon,
     TwitterIcon,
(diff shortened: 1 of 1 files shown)


manisandro:
Thanks - at this point it would make sense to already implement the configurability of the individual share buttons.

claudiadragoste:
Totally agree. Do you know where I can find something similar? I think it's easy to implement, with some example I could dare to implement it. Thank you

claudiadragoste:
Hey!! Thank you for these lines!! :) It's super nice like this.  I will think on it

NullSeile:
We've implemented a possible solution using the proposed changes. The modification has been done in this commit: b0d6a8b0cf3b960a2ad6d98dea3d396c97705ada.

manisandro:
Thanks! - Could you open a PR with that commit?

manisandro:
Addressed by #203

PR #206, opened 2022-06-07
add buffer unit

stephaschu:
Allows to change the buffer unit to metres, kilometres, feet and miles like in measurement plugin.


manisandro:
Thanks!

PR #207, opened 2022-07-15
Customizable ShareSocials component (#203)

NullSeile:
We've implemented a possible solution to #203 using the proposed changes.


manisandro:
Thanks!

PR #212, opened 2022-08-11
Fixed ShareSocials.jsx bug

NullSeile:
Fixed a bug created at #207 pull request where the url was not set correctly.


manisandro:
Thanks!

PR #215, opened 2022-10-04
fix vectorlayerutils recursive call

tobias-brunner:
Currently, the precision parameter of the geoJSONGeomToWkt Function is set to the index of the object inside the GeometryCollection when called recursively.


manisandro:
Thanks!

PR #216, opened 2022-10-05
Implement f url parameter for wms filter

raynerhoward:
This is my attempt at implementing a new qwc2 url parameter f that simply gets passed to wms getmap and getfeatureinfo requests.  Seems to work.  My first ever pull request, hope I'm doing this right :-)

### DIFF
--- a/components/map/layers/WMSLayer.js
+++ b/components/map/layers/WMSLayer.js
@@ -12,13 +12,17 @@ import url from 'url';
 import CoordinatesUtils from '../../../utils/CoordinatesUtils';
 import ConfigUtils from '../../../utils/ConfigUtils';
 import MapUtils from '../../../utils/MapUtils';
+import {UrlParams} from '../../../utils/PermaLinkUtils';
 
 
 function wmsToOpenlayersOptions(options) {
     const urlParams = Object.entries(url.parse(options.url, true).query).reduce((res, [key, val]) => ({...res, [key.toUpperCase()]: val}), {});
+    const urlFilter = UrlParams.getParams()['f'];
+
     return {
         ...urlParams,
         LAYERS: options.name,
+        FILTER: urlFilter,
         STYLES: options.style || "",
         FORMAT: options.format || 'image/png',
         TRANSPARENT: options.transparent !== undefined ? options.transparent : true,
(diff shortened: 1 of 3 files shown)


manisandro:
Thanks for working on this. Unfortunately this only covers a very narrow use-case, for instance it may also lead to undesired resp broken behaviour when multiple WMS are added to the map and a layer name specified in the filter expression does not exist in one of the loaded WMS layers, as such I don't believe it's safe to add to qwc2 upstream in this form.

One approach would be to extend the syntax of the layer entry in the `l` parameter. Currently, it is of the form

    <layername>[<transparency>]!

One could extend it to say

        <layername>[<transparency>]{<filter>}!

this way, you can specify a filter independently for each layer.

raynerhoward:
I like how I can copy the entire GetMap FILTER url parameter:

https://roadalignments.com/tdot/G03-TDOT?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image/png&TRANSPARENT=true&LAYERS=Grid,Boundary,Segments,Verticies&FILTER=Segments:"alid" = '8e9fd94d-3fbc-4ab3-8da9-9808af10e233';Verticies:"alid" = '8e9fd94d-3fbc-4ab3-8da9-9808af10e233';Boundary:"alid" = '8e9fd94d-3fbc-4ab3-8da9-9808af10e233';Grid:"alid" = '8e9fd94d-3fbc-4ab3-8da9-9808af10e233'&CRS=EPSG:2274&WIDTH=756&HEIGHT=901&BBOX=2173391.2655798886,261660.26880698328,2174626.1772201117,263132.03519301675

and paste it into my qwc2 'f' filter parameter:

https://roadalignments.com/tdot-qwc2/?t=G03-TDOT&f=Segments:"alid" = '8e9fd94d-3fbc-4ab3-8da9-9808af10e233';Verticies:"alid" = '8e9fd94d-3fbc-4ab3-8da9-9808af10e233';Boundary:"alid" = '8e9fd94d-3fbc-4ab3-8da9-9808af10e233';Grid:"alid" = '8e9fd94d-3fbc-4ab3-8da9-9808af10e233'&l=Verticies,Segments,Boundary,Grid&e=2173306.9779600003,261754.32515999998,2174710.46484,263037.97884

But yeah, I get what you're saying.

Regards

PR #217, opened 2022-10-19
Possibility for set a custom legend icon in layer tree

danceb:
I thought, it would be nice, to have the possibility to set a custom legend icon for each layer within the layer tree instead of the legend thumbnail.
So this is my attempt for realizing this.

### DIFF
--- a/plugins/LayerTree.jsx
+++ b/plugins/LayerTree.jsx
@@ -61,6 +61,8 @@ class LayerTree extends React.Component {
         setActiveServiceInfo: PropTypes.func,
         setSwipe: PropTypes.func,
         showLegendIcons: PropTypes.bool,
+        showCustomLegendIcon: PropTypes.bool,
+        iconFormat: PropTypes.string,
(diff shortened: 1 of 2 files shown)


manisandro:
This functionality is provided by the `qwc-legend-service` (https://github.com/qwc-services/qwc-legend-service), would you be ok with using the service or do you still prefer it to be in the client?

danceb:
Oh, I missed that! Of course, it is ok to use the service instead of a configuration within the client. So I will have a look at the service configuration now. Thanks for the hint!

danceb:
Just have to ask: It seems, that the service is for a very detailed configuration for each layer within a wms service. But is it possible to just provide one icon/image and that one is used on every layer? That was the thing I was looking for. Thanks!

manisandro:
Currently not, but should be pretty easy to implement here: https://github.com/qwc-services/qwc-legend-service/blob/master/legend_service.py#L278

PR #218, opened 2022-11-04
Refactor StandardApp component as function React component using hooks

benoitblanc:
Hi,

I suggest this PR to start using React function components and hooks from React >= v16.8. It could improve readibility and maintenability, and be familiar with some features of React newest versions.

Thanks

### DIFF
--- a/components/StandardApp.jsx
+++ b/components/StandardApp.jsx
@@ -6,9 +6,9 @@
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import React from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
-import {Provider, connect} from 'react-redux';
+import {Provider, useSelector, useDispatch} from 'react-redux';
 
 // Needed for IE11 to avoid 'Promise not defined' error in axios
 import "core-js/stable";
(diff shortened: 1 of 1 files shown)


manisandro:
Granted that I've so far not worked a lot with function components, personally I find them much less readable than class components. What other advantages do you see?

benoitblanc:
I agree that readable code is subjective.

Other advantages of using function components are:
* less code
* reusable code by using functions to extract component logic
* using hooks to deal with state, or create custom hooks for your own component logic without create a class and make your component Stateful
* there can be some performance optimization to avoid DOM rendering by using specific hooks `useMemo`, `useEffect`

It is not mandatory to have all of the components as function components, we can have both of function and class components. Here is an example that could be extended in the future to other existing components or new components.

We could also use [Redux Toolkit](https://redux.js.org/introduction/why-rtk-is-redux-today) instead of `react-redux` to make store, reducers, actions easier to use but it is an other topic to enhance qwc2 app. https://redux.js.org/introduction/why-rtk-is-redux-today

manisandro:
While I don't see any problem with starting to adopt function components in newly QWC plugins if the author wishes to use them and gradually extending them to more code if they bring a true benefit, I don't think that proactively touching existing code and risking subtle breakages is worthwhile, also considering the effort of reviewing all these changes. As far as I can see there are no indications that class components will be deprecated.

benoitblanc:
Yep I understand your point of view about breakages and review of these changes. I close this one ;-)

PR #223, opened 2022-12-13
fix small bug preventing to use displayfield

olivierdalang:
Fixes a minor typo (`displayField` instead of `displayfield`) that prevents QWC from using the display field configured in the QGIS project


manisandro:
Good catch, thanks!

PR #225, opened 2022-12-21
Update renderLayers in Map.jsx

rsrg-zwiama:
In the renderLayer function the zIndex is set in the loop starting from 0, incrementally increased and set as options.zIndex.  The layer.zIndex can be defined in MeasurementSupport.jsx or EditingSupport.jsx and others, but is not used to render the layers. So the layer order isn't influenced by setting layer.zIndex. Due to that, the marker-icon from height profile renders underneath the profile line.
I added a short if clause to use the value of layer.zIndex.

### DIFF
--- a/plugins/Map.jsx
+++ b/plugins/Map.jsx
@@ -179,6 +179,9 @@ class MapPlugin extends React.Component {
         return this.state.renderLayers.map(layer => {
             ++zIndex;
             const options = {...layer, zIndex: zIndex};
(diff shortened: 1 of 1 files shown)


manisandro:
Thanks! I've just applied a small syntax change.

rsrg-zwiama:
> Thanks! I've just applied a small syntax change.

thanks, elegant solution :-)

PR #226, opened 2023-01-13
Allow ResizeableWindow to be docked to any side

smaspons:
I needed it so maybe it's useful to someone else.

I'm thinking if it would be interesting to add an option to dock it to the full height, like it does when docking at the bottom but on the right/left too. 
I personally don't see why you'd want it to be resized when docked to the sides but not when docked to the bottom. If I wanted that I would just move it to the top-left corner of the map and leave it there, it would serve the same purpouse... To me if you dock it it should occupy all the space on that side.

I might be wrong so please, tell me what you think so we can implement this properly.
Thanks!

### DIFF
--- a/components/ResizeableWindow.jsx
+++ b/components/ResizeableWindow.jsx
@@ -150,10 +150,7 @@ class ResizeableWindow extends React.Component {
         const dockSide = this.props.dockable === true ? "left" : this.props.dockable;
         const dockIconSuffix = (dockSide === "bottom" || dockSide === "top") ? "_bottom" : "";
         let dockIcon = docked ? 'undock' : 'dock';
-        if (dockSide === "right" || dockSide === "top") {
-            dockIcon = docked? 'dock' : 'undock';
-        }
-        dockIcon = dockIcon + dockIconSuffix;
+        dockIcon = dockIcon + "_" + dockSide;
 
         const content = [
             (<div className="resizeable-window-titlebar" key="titlebar" onDoubleClick={this.toggleMaximize} ref={el => { this.titlebar = el; }}>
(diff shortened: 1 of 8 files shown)


components/ResizeableWindow.jsx
```diff
@@ -142,12 +142,18 @@ class ResizeableWindow extends React.Component {
...
         const zIndex = 10 + this.props.windowStacking.findIndex(item => item === this.id);
-        const dockBottom  = this.props.dockable === "bottom";
-        const dockIconSuffix = dockBottom ? "_bottom" : "";
+        let docked = this.state.geometry.docked;
+        const dockSide = this.props.dockable === true ? "left" : this.props.dockable;
+        const dockIconSuffix = (dockSide === "bottom" || dockSide === "top") ? "_bottom" : "";
+        let dockIcon = docked ? 'undock' : 'dock';
+        if (dockSide === "right" || dockSide === "top") {
+            dockIcon = docked? 'dock' : 'undock';
+        }
```
manisandro, line 155: I think it would be better to provide dedicated icons for dock/undock top/right.

smaspons: Yes, I did that in [this commit](https://github.com/qgis/qwc2/pull/226/commits/433f65077b953f7f2608da9620e3ad610f7acfb9)

smaspons:
Do you think that docked windows at left/right should occupy all the screen height? Just like docking at the bottom makes the window's width 100%

smaspons:
Yeah, I've been thinking about it and what you say is correct.
I think it can be merged then.

smaspons:
Yes, sorry, my editor auto formatted the file. Now it should be how it was before.

manisandro:
Thanks!

PR #227, opened 2023-01-20
Improve AttributeTable component

smaspons:
I added a prop to hide/show the "Open Edit Form" button. By default it's visible if the plugin Editing is enabled.

Also, I saw that when you use the "Zoom to feature" button with a single point selected, it uses the smallest scale avaliable. I added a prop to configure the zoom scale, and it only applies when zooming to a single point.

Please let me know if I did any mistakes, thanks!

### DIFF
--- a/plugins/AttributeTable.jsx
+++ b/plugins/AttributeTable.jsx
@@ -14,7 +14,7 @@ import isEmpty from 'lodash.isempty';
 import ContentEditable from 'react-contenteditable';
 import {getFeatureTemplate} from '../actions/editing';
 import {LayerRole} from '../actions/layers';
-import {zoomToExtent} from '../actions/map';
+import {zoomToExtent, zoomToPoint} from '../actions/map';
 import {setCurrentTask, setCurrentTaskBlocked} from '../actions/task';
 import EditComboField, {KeyValCache} from '../components/EditComboField';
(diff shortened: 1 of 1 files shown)


plugins/AttributeTable.jsx
```diff
@@ -545,7 +554,12 @@ class AttributeTable extends React.Component {
             features: this.state.filteredSortedFeatures.filter(feature => this.state.selectedFeatures[feature.id] === true && feature.geometry)
         };
         if (!isEmpty(collection.features)) {
-            this.props.zoomToExtent(geojsonBbox(collection), this.props.mapCrs);
+            if (collection.features.length == 1 && collection.features[0].geometry.type === "Point"){
```
manisandro, line 557: Does this work with a geometryless feature?

manisandro:
Looks good overall thanks!

plugins/AttributeTable.jsx
smaspons: I'm afraid not. But I could just use the zoomToExtent if it's a geometryless feature, right?

manisandro: Ah right the collection anyway only includes features with geometries, so all good.

manisandro:
Thanks!

PR #228, opened 2023-01-27
Identify: Use layer filters in GetFeatureInfo request

smaspons:
Hi,

I thought the Identify tool should only return the features you can see, so it should have the same filters as the layers. Please let me know if this is not the correct way to do it.

### DIFF
--- a/plugins/Identify.jsx
+++ b/plugins/Identify.jsx
@@ -99,7 +99,12 @@ class Identify extends React.Component {
             let queryableLayers = [];
             queryableLayers = IdentifyUtils.getQueryLayers(this.props.layers, this.props.map);
             queryableLayers.forEach(l => {
-                const request = IdentifyUtils.buildRequest(l, l.queryLayers.join(","), clickPoint, this.props.map, this.props.params);
+                const params = {...this.props.params};
+                // Apply layer filters to identify request
+                if (l.params.FILTER) {
+                    params.filter = l.params.FILTER
+                }
+                const request = IdentifyUtils.buildRequest(l, l.queryLayers.join(","), clickPoint, this.props.map, params);
                 ++pendingRequests;
                 IdentifyUtils.sendRequest(request, (response) => {
                     this.setState({pendingRequests: this.state.pendingRequests - 1});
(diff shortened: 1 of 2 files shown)


smaspons:
I tried doing it for the other modes (region & radius) but it doesn't work properly, it returns all the features. It seems that if I add the `filter` param it ignores the `FILTER_GEOM` param. 
Am I doing something wrong or the `filter` and `FILTER_GEOM` params are  mutually exclusive?

plugins/Identify.jsx
```diff
@@ -99,7 +99,12 @@ class Identify extends React.Component {
             let queryableLayers = [];
             queryableLayers = IdentifyUtils.getQueryLayers(this.props.layers, this.props.map);
             queryableLayers.forEach(l => {
-                const request = IdentifyUtils.buildRequest(l, l.queryLayers.join(","), clickPoint, this.props.map, this.props.params);
+                const params = {...this.props.params};
+                // Apply layer filters to identify request
+                if (l.params.FILTER) {
+                    params.filter = l.params.FILTER
+                }
+                const request = IdentifyUtils.buildRequest(l, l.queryLayers.join(","), clickPoint, this.props.map, params);
```
manisandro, line 107: Perhaps better to apply the layer filters inside IdentifyUtils.buildRequest?

smaspons: Yes, but as I said identifyRegion uses the param `FILTER_GEOM` and if I add the param `filter` it doesn't filter properly, it returns features from outside the identified region.
I don't really know what is happening, to me it seems like a qgis server thing, no?

manisandro: Yes I believe it is a QGIS Server limitation

smaspons: Okay, well, I will still move it inside IdentifyUtils.buildRequest as I see that it's only called for identifyPoint

manisandro: I believe the cause is this: https://github.com/qgis/QGIS/blob/master/src/server/services/wms/qgswmsrenderer.cpp#L1407

manisandro:
Thanks!

PR #230, opened 2023-02-03
Update spanish & catalan translations

smaspons:
Only one question, why is the "External Link" translation in qwc2-demo-app and not in qwc2?


manisandro:
Thanks - it's because the "External Link" menuentry is just an example entry in the demo app, it is not part of the qwc2 core.

PR #232, opened 2023-02-17
support for extra legend parameters

danceb:
add support for adding extra legend parameters to getLegendGraphic request as in MapLegend plugin

### DIFF
--- a/plugins/LayerTree.jsx
+++ b/plugins/LayerTree.jsx
@@ -43,6 +43,7 @@ class LayerTree extends React.Component {
         enableLegendPrint: PropTypes.bool,
         enableServiceInfo: PropTypes.bool,
         enableVisibleFilter: PropTypes.bool,
+        extraLegendParams: PropTypes.string,
         fallbackDrag: PropTypes.bool,
         flattenGroups: PropTypes.bool,
         grayUnchecked: PropTypes.bool,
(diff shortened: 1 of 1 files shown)


danceb:
After adding this parameter I recognized, that there is another theme specific parameter called `extraLegendParameters`, which is set within the themesConfig.json. Does this do the same like the parameter here? So maybe we should name it the same within the LayerTree and the MapLegend plugin? So it would be `extraLegendParameters` instead of `extraLegendParams` ?

manisandro:
Yes that might be a good idea, and since the MapLegend plugin is very fresh it should not cause breakage for many users

plugins/style/Buttons.css
```diff
@@ -16,7 +16,7 @@ button.map-button {
     cursor: pointer;
 }
 
-button.map-button:disabled {
+button.map-button.locate-button-DISABLED {
```
manisandro, line 19: This looks like an accidental commit?

danceb: Not really... I recognized that the map locator button didn't get transparent, if I block the location within the browser. Than I checked the css class of the button and it is named "locate-button-DISABLED". So I changed this within the Buttons.css.

manisandro: Ah ok!

manisandro:
Thanks!

plugins/style/Buttons.css
danceb: It hasn't anything to do with the original commit here, that's right. But as it is just a small change, I wouldn't start another pull request for it.

PR #235, opened 2023-04-06
Making ProcessNotifications draw over everything else

NullSeile:
We added a property to the `ProcessNotification` plugin. This option allows it to draw itself on top of all other plugins. This way we can avoid ResizeableWindows or SideBars hiding the notifications.


manisandro:
Thanks - I think you'll always want the notifications on top, so this could also just be hard-coded in ProcessNotifications.css

NullSeile:
Sure! Its hardcoded now in ProcessNotifications.css

manisandro:
Thanks

PR #236, opened 2023-04-13
themeBlacklist on menu items and toolbar items

NullSeile:
We added the ability to set a theme blacklist for menu items and toolbar items. It can coexist with the current theme whitelist.


manisandro:
Thanks

PR #237, opened 2023-04-14
TimeManager improvements

HusseinKabbout:
This PR introduces the following improvements:

1. Faulty datetime parsing

The following datetime strings are not parsed correctly at the moment:

- `2023-03-25 00:00:00 (UTC)`
- `2023-04-01 00:00:00 (UTC+02:00)`

This PR adds a new custom date format as specified in [adding-custom-formats](https://github.com/kensnyder/any-date-parser#adding-custom-formats) to parse them correctly.

2. QGIS server returns no features with `enddate === NULL`

At the time of writing, QGIS server does not return any feature that does not have `enddate` set. To workaround this limitation, a placeholder date should be used instead to make features with no `enddate` visible.

Proposed placehold: `9999-01-01 00:00:00`

### DIFF
--- a/plugins/TimeManager.jsx
+++ b/plugins/TimeManager.jsx
@@ -49,7 +49,7 @@ const DateUnitLabels = {
     "100y": LocaleUtils.trmsg("timemanager.unit.century")
 };
 
-let qgis_date_format = new Format({
+const qgis_date_format = new Format({
     //        $dateExpr           $hour        $minute         $second             $millisecond                 $zone                 $offset
     matcher: /^(.*?)[\s,-]*([01]\d|2[0-3])\:([0-5]\d)(?:\:([0-5]\d|60)(?:[\.,](\d{9}|\d{6}|\d{1,3}))?)?[\s,-]*\(?(UTC)?[\s,-]*([+-]0\d?\:?(?:[0-5]\d)?)?[\s,-]*\)?$/i,
     handler: function ([match, dateExpr, hour, minute, second, millisecond, zone, offset]) {
--- a/plugins/TimeManager.jsx
+++ b/plugins/TimeManager.jsx
@@ -78,6 +78,12 @@ const qgis_date_format = new Format({
 });
 dateParser.addFormat(qgis_date_format)
 
+// QGIS server does not return any feature that does not have "enddate" set.
+// To workaround this limitation, a placeholder date is used to make features
+// with no "enddate" visible. This variable represents that placeholder date.
+// This information is needed in the QWC2 so that features with no "enddate"
+// are represented correctly. It is also used to differentiate them from features with
+// a valid "enddate".
 const DUMMY_END_DATE = new Date('9999-01-01 00:00:00');
 
 /**
--- a/plugins/TimeManager.jsx
+++ b/plugins/TimeManager.jsx
@@ -78,6 +78,8 @@ const qgis_date_format = new Format({
 });
 dateParser.addFormat(qgis_date_format)
 
+const DUMMY_END_DATE = new Date('9999-01-01 00:00:00');
+
 /**
  * Allows controling the time dimension of temporal WMS layers.
  */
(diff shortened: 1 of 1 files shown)


plugins/TimeManager.jsx
```diff
@@ -191,11 +224,12 @@ class TimeManager extends React.Component {
                 }
             });
             timeData.values = [...timeData.values].sort().map(d => dayjs.utc(d));
+            const enddate = timeData.values.length > 0 ? timeData.values[timeData.values.length - 1].hour(23).minute(59).second(59) : null;
             this.setState((state) => ({
                 timeData: timeData,
                 currentTimestamp: state.currentTimestamp ?? +timeData.values[0],
                 startTime: timeData.values.length > 0 ? timeData.values[0].hour(0).minute(0).second(0) : null,
-                endTime: timeData.values.length > 0 ? timeData.values[timeData.values.length - 1].hour(23).minute(59).second(59) : null
+                endTime: enddate.year() !== 9999 ? enddate : null
```
manisandro, line 232: Rather than hardcoding `9999`, introduce a variable say `const DUMMY_END_DATE=9999;` at top level.

danceb:
Related to the behaviour of the QGIS Server regarding NULL values I have created an issue some time ago:
https://github.com/qgis/QGIS/issues/51246

The problem here occurs also, if no start time is set.

plugins/TimeManager.jsx
```diff
@@ -78,6 +78,8 @@ let qgis_date_format = new Format({
 });
 dateParser.addFormat(qgis_date_format)
 
+const DUMMY_END_DATE = new Date('9999-01-01 00:00:00');
```
manisandro, line 81: Can you add a quick comment why this is needed?

manisandro:
Thanks!

danceb:
But now features with no "start" value will not be visible yet, right? So in this case it is the same like with features which have no "end" value. So I would appretiate, if the code can be extended to pay attention on this case.

The extent (min, max) of the timeline is generated from min(start value) and max(end value) of the features within the map. Does the PR affects this, so the maximum of the timeline is loaded as '9999-01-01' and not the "right" maximum date? I would appretiate it as well, if you could pay attention on this case.

Thanks!

HusseinKabbout:
> But now features with no "start" value will not be visible yet, right?

I would argue that this would be the correct behavior. What is the use case of features with no start date? No end date means that the feature is currently still active. What would a missing start date mean for a feature?

> Does the PR affects this, so the maximum of the timeline is loaded as '9999-01-01' and not the "right" maximum date?

I am unsure if I understood you correctly. This PR sets the end date to `null` if it has the value '9999-01-01'. The `max` value that you speak of does not change, it is kept as '9999-01-01'. When the timeline is loaded we check for the `null` value and do the following:

- end date == `null` --> Set timeline extent to (start_date, '9999-01-01') and make initial zoom to (start_date,current_time)
-  end date != `null` --> Set timeline extent to (start_date, end_date) and make initial zoom to (start_date, end_date)

Does this answer your question?

danceb:
> What would a missing start date mean for a feature?

This would mean, that the feature is active before an end date. For instance in our specific case we have features of planned cunstruction projects for the future, where the end date is already set (the project has to be done until this date), but no start date is set yet, because it it is not clear at the moment, when the construction would start exactly. And this features are not be shown within the map viewer.

Related to this there is also the possibility, that no start date AND no end date is set. Using the time manager within QGIS this features are shown all the time because they are active regardless which date is selected on the timeline. QGIS server does not follow this, so these features wouldn't be visible as well. If the start date is considered here as well, we would see these features.

So for this case we need to specify a dummy value for the start date as well. Is `'0000-01-01'` a usable value for this?

> This PR sets the end date to null if it has the value '9999-01-01'

> end date == null --> Set timeline extent to (start_date, '9999-01-01')

I am sorry, but I don't understand this exactly. If the end date is `'9999-01-01'` within the data, it is set to `null` first and then the timeline extent is zoomed to (start_date, '9999-01-01')? But thats the problem I am asking for. So you will always have a timeline which is zoomed for the next 8k years to the future, which is not very handy. 

So lets summarize, I think this should be the procedure:
- set a dummy date within the date for start date AND/OR end date (that's what you are doing at the moment, right?)
- qgis server will render this features, because the dates are not `null`
- replace this values with `null` within the timemanager plugin
- for the display within the timeline specify, that: 
  - features with `start date == null and end date != null` --> active/visible before time marker
  - features with `start date != null and end date == null` --> active/visible after time marker
  - features with `start date == null and end date == null` --> active/visible before and after time marker
  - the timeline extent should be min/max of the features which does not have a dummy value respective `null`

Would this be a solution, where everyone is happy with?

danceb:
Hi @patricev, you are mixing two different things here: Displaying the features with NULL values within the qwc2 map viewer and displaying corresponding objects within the timetable view. Only the second one could be handled within the qwc2 application. If and how the features are diesplayed within the map canvas could only be handled be bringing the qgis server to show it the same way as it is in desktop qgis.

I will try to get more attention on the issue I have opened in the qgis repo, for instance with funding options.

PR #238, opened 2023-04-18
More TimeManager improvements

benoitblanc:
Hi,

I have tried to enable TimeManager by default with a new props in this PR. Also I wanted to set the plugin visible at the start of the application by setting `startupTask` in `config.json`.

I have used http://qwc2.sourcepole.ch/ows/observations for my tests and application crashed as there are some NULL values in `enddate` and `startdate` in some data.

I have also added a dummy value to handle start date. (ping @danceb)

### DIFF
--- a/plugins/TimeManager.jsx
+++ b/plugins/TimeManager.jsx
@@ -51,10 +51,10 @@ const DateUnitLabels = {
     "100y": LocaleUtils.trmsg("timemanager.unit.century")
 };
 
-const qgis_date_format = new Format({
+const qgisDateFormat = new Format({
     //        $dateExpr           $hour        $minute         $second             $millisecond                 $zone                 $offset
     matcher: /^(.*?)[\s,-]*([01]\d|2[0-3])\:([0-5]\d)(?:\:([0-5]\d|60)(?:[\.,](\d{9}|\d{6}|\d{1,3}))?)?[\s,-]*\(?(UTC)?[\s,-]*([+-]0\d?\:?(?:[0-5]\d)?)?[\s,-]*\)?$/i,
-    handler: function ([match, dateExpr, hour, minute, second, millisecond, zone, offset]) {
+    handler: function([match, dateExpr, hour, minute, second, millisecond, zone, offset]) {
         let result = {};
         if (dateExpr) {
             result = this.parser.attempt(dateExpr);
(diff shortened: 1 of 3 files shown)


benoitblanc:
Hi @manisandro @HusseinKabbout 

What do you think about this PR (kind of continuation of #237) ?

Thanks

HusseinKabbout:
@benoitblanc I wanted to take a look yesterday but got lost in work. The `enddate` dummy date change was maybe too early. I think it would be good to first check why QGIS server returns no features if either `startdate` or `enddate` are not set. If it is a bug then fixing QGIS server would the right way to fix our problem. But if this behavior is defined by the OGC standard then the dummy dates are the only workaround / fix. I will look into it in the next couple of days.

danceb:
AS mentiond in https://github.com/qgis/QGIS/issues/51246, I couldn't find a satisfying answer within the ogc wms specifications. Maybe you have another idea where to dig for information about this. 
Unfortunately there isn't anything to find about time dimension in WMS within the official QGIS documentation as well. The "TIME" paremeter is just missing there: https://docs.qgis.org/3.28/en/docs/server_manual/services/wms.html#getmap. Thanks in advance!

HusseinKabbout:
I (resp. @mhugent ) looked into it and it seems that the OGC standard does not define this behavior. That means that this is implementation specific behavior, which we can be changed. @mhugent and I will look into it and create a PR for QGIS server which should fix the "NULL" value problem.

benoitblanc:
Hi @manisandro ,

I saw that you have added some commits about TimeManager plugin. Is it possible to add (part of) commits from this PR too ?

* https://github.com/qgis/qwc2/pull/238/commits/8c435bb4aa9fd4531ebcdd5c5c0715d98b51ef43 to read `fieldName` and `endFieldName` when reading GetProjetSettings for themes generation
* https://github.com/qgis/qwc2/pull/238/commits/dbfa801195e8668fbe285c6dd0e678744e1ef7c7 to allow user to enable TimeManager by default
* https://github.com/qgis/qwc2/pull/238/commits/b77b21a0b1d22192ecf4b4bb73beca4e816709f2 for a possible crash when TimeManager is enabled by default (`timeData.layers` is not defined)

I think we could then close this issue as Null values will be handled in QGIS Server.

Thanks

manisandro:
Hi @benoitblanc, I've applied your commits, closing the PR. Thanks!

PR #240, opened 2023-05-17
Sidebar positioning through html classes and css

NullSeile:
Instead of setting the position of the sidebar in the .jsx file, it just sets the HTML classes. Then CSS will determine the positioning.

This change makes modifying the appearance of qwc2 much easier, without the need to modify the core code of qwc2.


manisandro:
Thanks!

PR #241, opened 2023-05-17
Adding a unique key to tiled layers to avoid cache

NullSeile:
We added the ability to set (and change) an extra parameter to the tiled requests url. This way, if tiles change, you can change this key so that the map updates correctly.

Right now it has been implemented only for WMTS layers.

### DIFF
--- a/components/map/layers/WMTSLayer.js
+++ b/components/map/layers/WMTSLayer.js
@@ -14,35 +14,52 @@ function getWMSURLs(urls) {
     return urls.map((url) => url.split("?")[0]);
 }
 
+function createWMTSSource(options) {
+    const urls = getWMSURLs(Array.isArray(options.url) ? options.url : [options.url]).map((url) => {
+        if (options.rev) {
+            return url + "&rev=" + options.rev
+        } else {
+            return url
+        }
+    });
+    const projection = ol.proj.get(options.projection);
+    const resolutions = options.resolutions;
+
+    const matrixIds = new Array(options.resolutions.length);
+    // generate matrixIds arrays for this WMTS
+    for (let z = 0; z < options.resolutions.length; ++z) {
+        matrixIds[z] = options.tileMatrixPrefix + z;
+    }
+
+    return new ol.source.WMTS({
+        urls: urls,
+        layer: options.name,
+        format: options.format,
+        projection: projection ? projection : null,
+        matrixSet: options.tileMatrixSet,
+        tileGrid: new ol.tilegrid.WMTS({
+            origin: [options.originX, options.originY],
+            resolutions: resolutions,
+            matrixIds: matrixIds,
+            tileSize: options.tileSize || [256, 256]
+        }),
+        style: options.style !== undefined ? options.style : '',
+        wrapX: options.wrapX !== undefined ? options.wrapX : true,
+        requestEncoding: options.requestEncoding !== undefined ? options.requestEncoding : "REST"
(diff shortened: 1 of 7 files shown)


NullSeile:
This may give some problems when refreshing the page, as the rev=0 images may still be in cache.

components/map/layers/WMTSLayer.js
```diff
@@ -14,35 +14,52 @@ function getWMSURLs(urls) {
...
-                style: options.style !== undefined ? options.style : '',
-                wrapX: options.wrapX !== undefined ? options.wrapX : true,
-                requestEncoding: options.requestEncoding !== undefined ? options.requestEncoding : "REST"
-            })
+            source: createWMTSSource(options)
         });
+    },
+    update: (layer, newOptions, oldOptions) => {
+        if (newOptions.rev !== oldOptions.rev){
+            layer.setSource(createWMTSSource(newOptions))
```
manisandro, line 62: Does `ol.source.WMTS.refresh()` work here?

manisandro:
> This may give some problems when refreshing the page, as the rev=0 images may still be in cache.

If we are dealing with the browser cache, you'll probably need to add a timestamp or similar to the URL anyway, merely recreating the source won't help. If we are dealing with the OpenLayers cache, it will get cleared anyway on a page reload.

NullSeile:
Oh yes, it's the browser cache that's giving us problems. That's why we added the URL parameter. 
As is it should work fine except for what I explained here
> This may give some problems when refreshing the page, as the rev=0 images may still be in cache.

To solve this while using the current rev system we need a way to change the rev value manually. Does this sound good to you?

manisandro:
Yes, or rather than a counter we could just always use a YYYYMMDD timestamp in rev always?

NullSeile:
Sure!

manisandro:
> Yes, or rather than a counter we could just always use a YYYYMMDD timestamp in rev always?

Probably more like a full unix timestamp in milliseconds through, no point having day resolution

NullSeile:
Done! It should be working now

components/map/layers/WMTSLayer.js
```diff
@@ -14,35 +14,52 @@ function getWMSURLs(urls) {
     return urls.map((url) => url.split("?")[0]);
 }
 
+function createWMTSSource(options) {
+    const urls = getWMSURLs(Array.isArray(options.url) ? options.url : [options.url]).map((url) => {
+        if (options.rev) {
+            return url + "&rev=" + options.rev
```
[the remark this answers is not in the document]
NullSeile: Yes, it does. It works fine in our case, but it may not be standard

components/map/layers/WMTSLayer.js
```diff
@@ -14,35 +14,52 @@ function getWMSURLs(urls) {
     return urls.map((url) => url.split("?")[0]);
 }
 
+function createWMTSSource(options) {
+    const urls = getWMSURLs(Array.isArray(options.url) ? options.url : [options.url]).map((url) => {
+        if (options.rev) {
+            return url + "?rev=" + options.rev
```
manisandro, line 20: I guess there is a slim chance that this may collide with a genuine query parameter of the service. I'd rather just pass the rev directly as a query key without value, which should still be a valid request, i.e.

                return url + "?" + options.rev

manisandro:
Thanks!

PR #242, opened 2023-05-25
Fix empty values

beeekey:


manisandro:
Hi, thanks for your contribution, but this PR doesn't look ready for merging?

danceb:
While using the qwc2 as an important tool, I don't get it, why many of these changes should affect all qwc2 users. What color the measurement is and which units could be used (I need more than only 'm') should be changed in your application only and not via a PR for all of us ...

manisandro:
Closing since this PR looks abbandoned.

PR #243, opened 2023-06-05
Handle features without geometry

NullSeile:
Transforming geometries in a VectorLayer crashes when the feature does not have one.


manisandro:
Thanks!

PR #245, opened 2023-06-15
Add format and requestEncoding parameters in WMTS config generator

benoitblanc:
Hi,

WMTS capabilities contain `format` and `requestencoding` parameters that are used to create our WMTS background layer configuration in `themesConfig`.

This PR adds these parameters in the output of `wmts_config_generator` script, as it was done in https://github.com/qwc-services/qwc-admin-gui/pull/45 some time ago.

Thanks


manisandro:
Thanks

PR #246, opened 2023-06-19
Added IdentifyPoint translations text

NullSeile:
We added text for the IdentifyPoint app menu item, currently only in Catalan, Spanish, and English.


manisandro:
Thanks!

PR #247, opened 2023-07-04
Style marker and edit

ger-benjamin:
For https://github.com/qgis/qwc2-demo-app/issues/462

Based on the utils/FeatureStyles and the config files mechanism:
 - Add Styling possibility for the marker via a new `defaultMarkerStyle` config key.
 - Moves Measure/snap/edit/redlining styling to FeatureStyle.
 - Add Styling possibility to Measure/snap/edit/redlining via a new `defaultInteractionStyle` config key.
 - Add missing modify style on redlining and edition and link it to the existing measurement modify style (now under the `DEFAULT_INTERACTION_STYLE.sketch...` keys).

I tried to keep every differences between edit/snap/measurement/redlining styles. Hope it's fine, let me know if I've to rework something.

Related PR:
 -  https://github.com/qwc-services/qwc-services.github.io/pull/10 (doc)
 - (no default config to add somewhere ?)

### DIFF
--- a/plugins/map/EditingSupport.jsx
+++ b/plugins/map/EditingSupport.jsx
@@ -11,6 +11,7 @@ import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
 import ol from 'openlayers';
 import {setEditContext} from '../../actions/editing';
+import FeatureStyles from '../../utils/FeatureStyles';
 
 class EditingSupport extends React.Component {
     static propTypes = {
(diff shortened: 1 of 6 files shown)


plugins/map/EditingSupport.jsx
```diff
@@ -24,38 +26,23 @@ class EditingSupport extends React.Component {
...
-                    } else if (feature.getGeometry().getType() === "MultiPolygon") {
-                        return new ol.geom.MultiPoint(feature.getGeometry().getCoordinates()[0][0]);
-                    } else {
-                        return feature.getGeometry();
-                    }
-                }
-            })
-        ];
+        this.editStyle = FeatureStyles.interaction();
+        this.editStyle[1].setGeometry((feature) => {
```
manisandro, line 30: This is probably a bit hard to understand resp it requires you to know about the implementation detail of `FeatureStyles.interaction()`. It might be better to passs `geometry` as an argument to `FeatureStyles.interaction()` and then set the geometry accordingly inside `FeatureStyles.interaction()`.

plugins/map/MeasurementSupport.jsx
```diff
@@ -181,33 +167,16 @@ class MeasurementSupport extends React.Component {
...
-                    } else if (f.getGeometry().getType() === "LineString") {
-                        return new ol.geom.MultiPoint(f.getGeometry().getCoordinates());
-                    } else {
-                        return new ol.geom.MultiPoint(f.getGeometry().getCoordinates()[0]);
-                    }
-                }
-            })
-        ];
+        const styles = FeatureStyles.measureInteraction(feature);
+        styles[1].setGeometry((f) => {
```
manisandro, line 171: Same comment as above regarding `FeatureStyles.interaction()`.

plugins/map/EditingSupport.jsx
ger-benjamin: You're right, thanks. It's updated.
I can also split the first and the second style from "interaction" to be more clear if needed.

manisandro: Yes I think it would be better to split the interaction styles. Thanks.

manisandro:
Thanks!

ger-benjamin:
You're fast! Thanks to you :yellow_heart:

PR #249, opened 2023-07-10
Use new URL scheme

otbutz:
See https://community.openstreetmap.org/t/a-b-c-tile-openstreetmap-de-subdomains-von-tile-openstreetmap-de-werden-aufgehoben/100830/2

### DIFF
--- a/utils/ConfigProvider.js
+++ b/utils/ConfigProvider.js
@@ -15,7 +15,7 @@ export default {
                 }
             },
             DE: {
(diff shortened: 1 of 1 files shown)


manisandro:
Thanks

otbutz:
@manisandro is there a reason why all URLs default to http instead of https ? This would require at least one additional round trip to be informed about the permanent redirect.

manisandro:
Actually no, would make sense to change it

PR #251, opened 2023-07-26
bring window geometry configuration to all plugins with dockable window

danceb:
In some plugings it was already possible to specify the geoemtry of the dockable window with `initialWidth`, `initialHeight`, `initialX`, `initialY` and `initiallyDocked`.
This pull request brings this configuration possibility to all the other plugins with dockable windows and match the parameter names.


manisandro:
I had also though about such a change in the past but was reluctant to possibly break users configurations, but maybe it is better to change it after all and document the breaking change.

danceb:
There were a few times in the past, where I have to recheck and compare my configurations with the current qwc2 and the demo-app to get it work again (parameter and path changes for instance). So I think, it wouldn't be a huge problem, if this is mentioned within the change notes. Especially with the detailed plugin documentation, where the changes should be inserted automatically, right?

manisandro:
Thanks, merged, will document the breaking change in the upgrade notes.

PR #252, opened 2023-07-31
Dynamic plugins for mapinfotooltip

rafaelnavarro03:
Be able to add plugins to the tooltip and there is no need to change this file, in the appConfig file the user must add his plugin to the list of plugins and it will be automatically added.
At the moment, I have only made a list of 3 props that can be sent that are configured in the plugin's cfg export. This list can be expanded.

### DIFF
--- a/plugins/MapInfoTooltip.jsx
+++ b/plugins/MapInfoTooltip.jsx
@@ -56,16 +56,16 @@ class MapInfoTooltip extends React.Component {
         plugins: {}
     };
     state = {
-        coordinate: null, elevation: null, extraInfo: null, newPoint: null
+        point: null, elevation: null, extraInfo: null
     };
     componentDidUpdate(prevProps) {
-        if (!this.props.enabled && this.state.coordinate) {
+        if (!this.props.enabled && this.state.point) {
             this.clear();
             return;
         }
         const newPoint = this.props.map.click;
         if (!newPoint || newPoint.button !== 2) {
-            if (this.state.coordinate) {
+            if (this.state.point) {
                 this.clear();
             }
         } else {
--- a/plugins/MapInfoTooltip.jsx
+++ b/plugins/MapInfoTooltip.jsx
@@ -32,6 +32,16 @@ import './style/MapInfoTooltip.css';
  * 
  * If `mapInfoService` in `config.json` points to a `qwc-mapinfo-service`, additional
  * custom information according to the `qwc-mapinfo-service` configuration is returned.
+ * 
+ * To add external plugins to the tooltip, include your plugin in the `plugins` array
+ * inside the `MapInfoTooltip` section of the `AppConfig` file.
+ * 
+ * Example:
+ * ```json
+ * {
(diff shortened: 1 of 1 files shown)


rafaelnavarro03:
Here is a sample plugin [SamplePlugin](https://gist.github.com/rafaelnavarro03/55a79cb72caab6f2b9f3b84f5ba29bf8)
To change initial variable by API calls they must be put in the **componentDidMount** function or **constructor**.

Add plugin to the appConfig file
`MapInfoTooltipPlugin: MapInfoTooltipPlugin({
                PluginSample: PluginSample
            }),`
If you need anything more, please ask me.

plugins/MapInfoTooltip.jsx
```diff
@@ -181,19 +183,37 @@ class MapInfoTooltip extends React.Component {
...
+                                    removeCoordinates: this.clear,
+                                    coordinates: this.state.coordinate,
+                                };
+                                Object.entries(ComponentRender.cfg).map(([key, value]) => {
+                                    if (value){
+                                        componentProps[key] = propsMapping[key];
+                                    }
+                                });
+                                ComponentRender = ComponentRender.controls
+                            }
```
manisandro, line 200: This part is really hard to understand, what is cfg?

rafaelnavarro03:
I unified newPoint and coordinates into point. My intention with the cfg was to make a system that in each plugin you can specify which parameters are passed to it. I don't know at which point passing unwanted parameters can affect in rendiment but since we are using here only 2 props is a better and simpler approach the one you suggested. Thanks for pointing it. For the plugin itself now there is no need to put cfg or controls just a normal export.

plugins/MapInfoTooltip.jsx
```diff
@@ -45,16 +45,18 @@ class MapInfoTooltip extends React.Component {
...
         cooPrecision: 0,
         degreeCooPrecision: 4,
         elevationPrecision: 0,
-        includeWGS84: true
+        includeWGS84: true,
+        plugins: {}
     };
     state = {
-        coordinate: null, elevation: null, extraInfo: null
+        coordinate: null, elevation: null, extraInfo: null, newPoint: null
```
manisandro, line 59: drop `coordinate`, `point` instead of `newPoint`

rafaelnavarro03:
I've changed the old `coordinate ` values with the new `point` one

manisandro:
Last round of comments:

- Please add some minimal documentation in the top docstring above the MapInfoTooltip class.
- `plugins` is passed as an object, but only it's values are ever used. Possibly better to directly pass an array, so that the order of the plugins is also 100% defined.
- There is a whitespace change, please avoid those.

rafaelnavarro03:
The appConfig now looks like this, the structure is in the doc
`MapInfoTooltipPlugin: MapInfoTooltipPlugin([
                Plugin1, Plugin2
            ]),`

manisandro:
I've merged this manually with some documentation improvements. Thanks for your contribution!

PR #253, opened 2023-08-08
Search provider text update in Searchbox

cazitouni:
Actually, when I click on a recent search it updates the text in the Searchbox.  This PR aims to implement the same logic for the searchproviders results.  It would be a great addition for the UX to do so.

I made a new function "updatetext" because I do not want the layer to get unselected if I click again, but I still want it to get unselected If I remove text.

I Don't know if my logic for this is good, do not hesitate to inform me If I need to change anything. 

Regards, Clément.

### DIFF
--- a/components/SearchBox.jsx
+++ b/components/SearchBox.jsx
@@ -188,7 +188,7 @@ class SearchBox extends React.Component {
                             {!this.isCollapsed(sectionId) ? (
                                 <div className="searchbox-results-section-body">
                                     {group.items.map((entry, idx) => (
-                                        <div className="searchbox-result" key={"c" + idx} onClick={() => {this.selectProviderResult(entry, provider); this.blur(); }} onMouseDown={MiscUtils.killEvent}>
+                                        <div className="searchbox-result" key={"c" + idx} onClick={() => {this.selectProviderResult(entry, provider); this.blur(); this.updateSearchText(null,  entry.label || entry.text)}} onMouseDown={MiscUtils.killEvent}>
                                             <span className="searchbox-result-label" dangerouslySetInnerHTML={{__html: entry.text.replace(/<br\s*\/>/ig, ' ')}} title={entry.label || entry.text} />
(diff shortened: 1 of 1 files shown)


cazitouni:
Hello @manisandro,

I allow myself to soliscit you on that.
Any feedback ? 

Thansk you in advance, kind regards, Clément.

manisandro:
Sorry for the late reply. Hmm perhaps I'm missing something, but as I read the code this would immediately re-trigger a new search with the text of the selected result as soon as a result is selected?

Note also that selecting a solr provider result also won't currently update the text in the search box. It is only selecting a recent search which does so.

cazitouni:
Hello, 

Thanks for the feedback. 
Yes It should re-trigger a search to filter with the text you clicked on, you still have a small delay on it (Like the orgininal search).
Like this, the search panel get closed and you are centered on your result but when you re-open the search box it's already pre-filtered. 
It should also be possible to just bring the text up without retrigerring a search but it'll feel strange in my opinion. 

For solr I wasnt able to test it because I didn't have a solr instance for now, but If it is necessary I can configure one. 

Regards, Clément.

manisandro:
> It should also be possible to just bring the text up without retrigerring a search but it'll feel strange in my opinion.

Why? As I see it (if the search provider is sane) it would just return the same result again as the one selected. Instead of re-triggering the search you could just immediately set the result list to the one and only selected result.

The one downside I see with this approach is that say you enter a broad search term (say "post office"), and you want to browse through the list to see the results on the map, you'd have to re-type the broad term every time.

cazitouni:
Ok I understand, I modified the commit. 
So something like this, that just update the text without retrigering a search could be better ? 

I do that because we got many user feedbacks who feel strange about not getting the text back in the bar when a search item is clicked.

thank you.

manisandro:
I've implemented this in https://github.com/qgis/qwc2/commit/09b8101ea0121585f243d0e84133fbcca5e39620, please give it a try.

cazitouni:
@manisandro, Works great thanks. Sorry my solution wasn't good enough to make it !

PR #254, opened 2023-08-20
Update czech translations

OndraM:
Hi, this is an partial update of Czech translations. I translated only those I use and can see in the UI, so I could validate the contextual meaning.


tpo:
@manisandro or @pka - from a quick diagonal reading: the translations look fine, if they're technically sound (no syntax errors or such), then I think merging is fine

PR #257, opened 2023-08-28
Make inital radius of Identify plugin configurable

OndraM:
Currently the Identify radius default is always 50 m, which is not quite flexible, e.g. if the map is usually viewed in a much lower scale.

So I propose this update.


manisandro:
Thanks!

PR #258, opened 2023-08-29
Fix incorrect title of terms window

OndraM:
Terms iframe window incorrectly uses label of `viewertitle_label`.


manisandro:
Good catch, thanks!

PR #259, opened 2023-08-29
Icons of inline window opened from TopBar and BottomBar

OndraM:
Currently, when iframe inline window is opened from either TopBar or BottomBar, they won't get any icon as other windows do. 

For example layers catalog window does have icon:
obrazek

However custom iframe windows won't get any icon:
obrazek

This PR reuses the (already required) `icon` of `menuItems` in TopBar as icon of the window.

```js
"menuItems": [
    {
        "key": "ExternalLink",
        "icon": "link", // this icon was already required, but was only used in the menu, now it is also passed to the window
        "url": "http://example.com",
        "target": "iframe"
    }
]
```

obrazek

It also adds special keys (`termsUrlIcon` and `viewertitleUrlIcon`) to define icons of links also in BottomBar configuration.

What do you think?

### DIFF
--- a/components/AppMenu.jsx
+++ b/components/AppMenu.jsx
@@ -191,7 +191,7 @@ class AppMenu extends React.Component {
             this.toggleMenu();
         }
         if (item.url) {
-            this.props.openExternalUrl(item.url, item.target, LocaleUtils.tr("appmenu.items." + item.key));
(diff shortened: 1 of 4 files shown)


manisandro:
Thanks, looks good!

PR #260, opened 2023-08-29
Make icon of resizable window large, so it is the same size as sidebar window icon

OndraM:
Currently, icons of resizable windows are smaller (14px) than icons of sidebar windows (21px), which look optically uneven:

obrazek

This PR changes icon size of resizable windows to large, the same size as is icon in sidebar window:

obrazek


manisandro:
Thanks!

PR #262, opened 2023-10-08
Romanian translation of the strings

TNick:
There are a lot of empty strings in the `ro-RO` file so I filled those up. Some minor changes to existing strings, too.


manisandro:
Thanks

PR #264, opened 2023-10-12
Account for file created by yarn install in gitignore

TNick:
On a fresh `git clone` followed by `yarn install` there are some files that are not excluded by `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,8 @@
 node_modules/
 npm-debug.log
+
+# On a fresh install with yarn 3.3.0 these extra files were generated.
+.yarn
+.pnp.cjs
+.pnp.loader.mjs
+yarn.lock


manisandro:
Did you run `yarn install` in the parent qwc2-app folder or in the `qwc2` submodule folder?

TNick:
I have cloned this repository and run „yarn install„ inside. So without the demo app.

To reproduce:

```bash
git clone https://github.com/qgis/qwc2.git
cd qwc2
yarn install
```

manisandro:
That is somewhat atypical as this repo is really designed to be used as a submodule / workspace folder, indeed it does not provide any webpack configuration to build an application.

TNick:
Well, in my limited experience that is also atypical. :)
Would you not be opened in the future to accept a pull request for a webpack configuration for building the library and to publish it under https://www.npmjs.com/? By adding it the user can still make use of it as a submodule but it would also allow the user to `yarn add qwc2`.

TNick:
And, while I have your attention, [.gitmodules](https://github.com/qgis/qwc2/blob/master/.gitmodules) serves no purpose. Ok to remove it in another pull request?

manisandro:
I have no objections to adding it to npm, we'd just need to make sure that the various scripts in the parent package.json still work.

.gitmodules indeed is a leftover, happy to remove it.

TNick:
Okay.

Sorry for leaving this here but the Issues are deactivated.

I plan on digging into the library this weekend. As a way to better understand it I will be writing tests and adding descriptive comments in the source code. Any thoughts on the test framework, comment style, so that the pull request is acceptable?

manisandro:
You can file issues in the qwc2-demo-app repository.

I don't have any strong preference for a test framework. Regarding comments it might be interesting to have a general developers documentation to be published to qwc-services.github.io/.

TNick:
Well, the test configuration will be hosted in this repository (I don't think it makes sense to have the test config in the demo app, but we can add a config there, too), so what about this pull request?

manisandro:
No objections to merging it, I was just trying to understand the rationale for running yarn install in the submodule folder, which I take now is for creating a npm module.

PR #266, opened 2023-10-15
Add tests, docs, build

TNick:
This is just a preview to get the discution started.

I has to upgrade `ol` and `ol-ext` to be able to run `jest`.


manisandro:
If I may add one remark right away, I'm a bit skeptical about comments just for the sake of writing a comment but which don't add any real information. In my view they just make the code more cluttered and ultimately less readable.

TNick:
Thanks for the feed-back. Do you have specific examples? 

Most of the commetns are written so that they show up in the documentation. I'm in the habbit of interseeding comments before every few lines of code because I find it easier to understand said code a few month after. Especially so with JS. Python is easier on the eyes. My personal experience with open source projects is that the contributors are more likely to contribute if the code is well commented.

That being said I admit that the main reason for writing the comments in code was to give me time to absorb the concepts and to form a mental map of the code.

manisandro:
For instance this comment does not really add any meaningful information further than what is already clear from the code:

image

TNick:
Is this the intendend behaviour? Only show the notification if the visibility of a different layer is set.

See the test here:
https://github.com/AdvaitaTopCad/qwc2/blob/8a49f2839ee67107be99a90fbf47d5c83adaaf91/utils/ThemeUtils.test.js#L243

manisandro:
Hi @TNick , having reorganized the qwc2 repos, I'm nearly ready to publish qwc2 to npm. I see there is a previous build of yours [1] https://www.npmjs.com/~tnick, would you mind removing it to free up the qwc2 name? Thanks!

TNick:
Done

manisandro:
The package is now published on [npm](https://www.npmjs.com/package/qwc2). I'm closing this PR, but feel free to reopen a new one if you want to resume work on testing etc.

PR #267, opened 2023-10-24
Its just me or this is useless?

TNick:
`if (newLayer.sublayers) {` is only true if `newLayer.sublayers === []` because of the ` if (!isEmpty(layer.sublayers)) {` above it.


manisandro:
This is needed to prevent a invariant state mutation.

TNick:
Sure, but would it not be the same as `newLayer.sublayers = []` or am I missing something from the logic of the code?

manisandro:
Sorry I didn't read this properly, you are right, thanks!

PR #268, opened 2023-10-26
Insert when a layer attribute is not found is being silently ignored.

TNick:
Either this or a console warning / assertion. It is used by the `addLayer` action, so maybe throwing an error is better?

### DIFF
--- a/utils/LayerUtils.js
+++ b/utils/LayerUtils.js
@@ -469,7 +469,10 @@ const LayerUtils = {
         if (index !== -1) {
             exploded.splice(index, 0, ...explodedAdd);
         } else {
-            exploded.splice(exploded.length, 0, ...explodedAdd);
+            throw new Error(
+                "Failed to find 'before' layer item with " + 
+                `'${beforeAttr}'=${beforeVal}`
+            );
         }
         return LayerUtils.implodeLayers(exploded);
     },
--- a/utils/LayerUtils.js
+++ b/utils/LayerUtils.js
@@ -468,6 +468,8 @@ const LayerUtils = {
         const index = exploded.findIndex(entry => entry.sublayer[beforeattr] === beforeval);
         if (index !== -1) {
             exploded.splice(index, 0, ...explodedAdd);
+        } else {
+            exploded.splice(exploded.length, 0, ...explodedAdd);
         }
         return LayerUtils.implodeLayers(exploded);
     },


TNick:
After writing the tests I see that the insert may be deep inside the tree, so it does not makes sense to simply append the layer.

manisandro:
This exception will though likely remain uncaught?

TNick:
It will remain uncaught because we want to see errors early, right? The `addLayer` function checks for this condition, so current implementation has no execution path that will throw an exception while running the application. Right now it only benefits the tests and protects future user from using this function in an unintended way.

TNick:
```
    /**
     * Inserts a layer into a tree.
     * 
     * The function creates a linear representation of the tree of layers
     * through {@link LayerUtils.explodeLayers}, inserts the layer
     * then it recreates the tree through {@link LayerUtils.implodeLayers}.
     * 
     * To determine the position of the insertion the function compares the
     * value of the `beforeAttr` property of each leaf layer with the
     * `beforeVal` argument.
     * 
     * @param {LayerData[]} layers - the list of layers to change
     * @param {LayerData} newLayer - the layer to insert
     * @param {string} beforeAttr - the attribute to examine (e.g. 
     *  `name` or `id`)
     * @param {*} beforeVal - the value to examine
     * 
     * @throws {Error} if the reference leaf layer is not found
     * @returns {LayerData[]} a new list that includes the `newLayer`
     */
```

This is the current documentation in the other pull request for the function, so the behaviour is documented.

TNick:
Assuming, of course, that that pull request will eventually be merged, of course. :)

manisandro:
I think it's a bad idea to expose any unhandled errors to the outside world, especially in a function indirectly called by a reducers, which can lead to the entire application crashing.

TNick:
That is a wise strategy when you have no tests in your code base.
I'm more of [fail fast](https://stackoverflow.com/a/2807375/1742064) camper.
Feel free to close this pull request if that is your final word.

manisandro:
This can depend on external input see (API.addExternalLayer) which in turn can depend on the environment in which the application runs, anche hence not testable. Since an uncaught exception can pull down the entire application and result in loss of work, this is a no-go.

TNick:
Okay, just leaving a final note for the future reader: `addExternalLayer()` uses `addLayer()` to insert the layer. `beforeLayerName` becomes `beforename` which is checked inside `ADD_LAYER` reducer, so the above argument has no legs.

Side note: the reducer should only be responsible for updating the store, everything else should be a [listner](https://redux.js.org/usage/side-effects-approaches).

manisandro:
Take
```
function addlayer() {
    window.qwc2.addExternalLayer("wms:https://wms.geo.admin.ch#ch.are.bauzonen", "foobarbaz");
}
```
where `foobarbaz` clearly does not exist. A thrown exception in `LayerUtils.insertLayer` is in this case only caught by chance in `ServiceLayerUtils.findLayers` which results in the (misleading) console error message

    Failed to read https://wms.geo.admin.ch?service=WMS&request=GetCapabilities

TNick:
Ok, I see your point, but surely ignoring the error cannot be the resolution.

I'm not sure that I understand the architecture, being stuck in `utils` and all, but if `addExternalLayer` is somehow at the border, it should thoroughly check all of its arguments.

Then again, when you say 

> an uncaught exception can pull down the entire application and result in loss of work, this is a no-go.

you're referring to the final user. It is not the final user that provides the `beforeLayerName` but - potentially - some user of this library. In which case the responsibility of catching the exception is with that code.

Bottomline, as I understand the code right now, the `beforeLayerName` parameter will not come from some user entering some text, but will be provided by the UI. If that is the case, an exceptions makes sense. If the input comes from the user, then it should be promoted to a visible notification (taking a hint here from your recent cange to the `master` branch).

manisandro:
Sure, one could have `addExternalLayer` check if the specified layer actually exists and return false/raise/watever, no issues with that. I just don't like the idea of raising exceptions deep down in the code, which is called through multiple paths, with no idea if and how these exceptions will be caught. I like the Qt approach here which is basically exception-less.

PR #269, opened 2023-10-27
Visibility is boolean

TNick:
Visibility is a boolean value in most (all?) other places. 
`visibilities` is a number array and is being checked below for equality with `0` and `0.5`.

### DIFF
--- a/utils/LayerUtils.js
+++ b/utils/LayerUtils.js
@@ -217,7 +217,7 @@ const LayerUtils = {
                 layernames.push(layer.type + ':' + (layer.capabilitiesUrl || layer.url) + "#" + layer.name);
                 opacities.push(layer.opacity);
                 styles.push(layer.style);
-                visibilities.push(layer.visibility);
+                visibilities.push(layer.visibility ? 1 : 0);
             } else if (layer.role === LayerRole.USERLAYER && layer.type === "separator") {
                 layernames.push("sep:" + layer.title);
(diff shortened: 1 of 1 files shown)


TNick:
Also, `buildWMSLayerUrlParam` makes no use of the `styles` array.

manisandro:
`visibilities` should be guaranteed to be set by the `ADD_LAYERS` reducer. This here

    if (visibilities[idx] === 0) {

looks like a potential bug however, probably the easiest fix would be to change this to

    if (visibilities[idx] === 0 || visibilities[idx] === false) {

? This should ensure the correct behaviour regardless of whether visibilities is set or not.

`styles` is only populated because `collectWMSSublayerParams` needs it, which is also called by other functions. Hypothetically at some point if QWC2 were to support switching layer styles, they could also be tracked in the URL.

TNick:
`visibilities` is a local variable. For theme layers is populated by `collectWMSSublayerParams` which does
```js
visibilities.push(layerVisibility ? (parentVisibility ? 1 : 0.5) : 0);
```

That's why I assumed that `0, 0.5, 1` is the set of allowed values.

`visibility` on the other hand may be the one you're thinking about.

manisandro:
I've pushed https://github.com/qgis/qwc2/commit/9b1fbf142cd79d16b100e192e0f9cba3aa5ef764, the rest should be correct.

PR #271, opened 2023-11-03
replace int id with uuid for kml import

cazitouni:
Right now, KML import uses integer types for IDs, which works but creates an error when you try to edit the layer with redlining. All elements disappear. Using a UUID, like the GeoJSON imports, seems to work better.


manisandro:
Thanks

PR #272, opened 2023-11-06
add possibility to not show the theme filter field within the theme switcher

danceb:
Filtering the themes is a useful feature if there are a lot of themes within the application but maybe not that necessary if there are only a few ones. So this PR adds the adds possibility to not show the theme filter field within the theme switcher via a new `showThemeFilter` parameter, which is `true` as default.


manisandro:
Thanks

PR #273, opened 2023-11-06
Add projection to MapInfoToolTip

cazitouni:
It could be interesting to pass the projection along with the point to the Tooltip plugin. This way, it's possible to perform transformations into different EPSG using CoordinatesUtils.reproject.

Doing this could allow for  interfaces between the tooltip and external applications that may expect a different projection than the one used in the map.

What do you think about this idea?

Thanks and kind regards,
Clément.


manisandro:
LGTM, thanks

PR #274, opened 2023-11-07
Also close 'failure' process notification

jblanchg:
I think it's a good idea to also close the process notification when the status is 'failure'. This way we avoid having a stack of multiple "failure" notifications on screen if, for some reason, there is multiple of them in a short period of time.


manisandro:
Why not, in case one can add a config option later on.

PR #277, opened 2023-11-15
Fix print download on chrome when using inlinePrintOutput

cazitouni:
Should fix https://github.com/qgis/qwc2-demo-app/issues/504.
Didn't find solution for the filename but, since you get the correct .pdf and you save it you can change the name anyway so maybe it's not that bad.

Let me know if it works for you.
I've tested to some extent with the demo project on firefox and chrome but not with atlas.

Kind regards,
Clement.

### DIFF
--- a/plugins/Print.jsx
+++ b/plugins/Print.jsx
@@ -232,7 +232,6 @@ class Print extends React.Component {
             <div className="print-body">
                 <form action={this.props.theme.printUrl} method="POST"
                     onSubmit={this.print} ref={el => { this.printForm = el; }}
-                    target="print-output-window"
                 >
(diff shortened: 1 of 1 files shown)


plugins/Print.jsx
```diff
@@ -545,7 +544,7 @@ class Print extends React.Component {
     print = (ev) => {
         if (this.props.inlinePrintOutput) {
             this.setState({printOutputVisible: true, outputLoaded: false});
-        } else {
+        }
```
manisandro, line 547: Can you fix the indentation of the following lines?

manisandro: The subsequent lines should be de-indented as the indentation block was removed.

cazitouni: MB sorry, should be fixed

manisandro:
Thanks!

PR #278, opened 2023-11-28
Do not reproject geometry for GeoJson format from GetFeatureInfo resp…

benoitblanc:
…onse as it is not necessary EPSG:4326 anymore

Hi @manisandro 

I'm not sure about this one... Maybe it breaks something elsewhere...

The problem is when we identify objects from external WMS added to the map, the geometry is not highlighted because there is a reprojection for GeoJSON format from EPSG:4326 to the map projection.

GetFeatureInfo from theme. Geometry is highlighted:

image

Import layer as WMS, and identify object. Geometry is not highlighted:

image

Without reprojection from 4326 to the map projection, geometry is highlighted because response is well parsed (it has already the right projection from request parameter):

image

GetFeatureInfo geometry response for geojson format is not necessary EPSG:4326 anymore. It can be other CRS, see this PR: https://github.com/qgis/QGIS/pull/32386

Thanks

### DIFF
--- a/utils/IdentifyUtils.js
+++ b/utils/IdentifyUtils.js
@@ -314,14 +314,10 @@ const IdentifyUtils = {
             if (result[layer.name] === undefined) {
                 result[layer.name] = [];
             }
-            let geometry = feature.geometry;
-            if (geometry) {
-                geometry = VectorLayerUtils.reprojectGeometry(geometry, "EPSG:4326", geometrycrs); // GeoJSON always wgs84
-            }
             result[layer.name].push({
                 ...feature,
                 id: id,
-                geometry: geometry,
+                geometry: feature.geometry,
                 layername: layer.name,
                 layertitle: layer.title
             });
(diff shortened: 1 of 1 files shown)


manisandro:
Does the GeoJSON feature response contain any CRS information?

benoitblanc:
No it does not...

manisandro:
Is this a GeoJSON response returned by QGIS Server? If yes, wouldn't it be better to extend QGIS Server to add a CRS property like
```
  "crs": {
    "type": "name",
    "properties": {
      "name": "urn:ogc:def:crs:XYZ
      }
    }
```
?

manisandro:
@benoitblanc According to [1] the Projection should now be read from `feature['crs']['properties']['name']` with fallback to `EPSG:4326`, right?

[1] https://github.com/qgis/QGIS/pull/56600

benoitblanc:
Hi @manisandro 

Sorry for the delay, I have completed the PR by reading CRS from GetFeatureInfo response (QGIS >= 3.38) and reproject geometry according to it.

Thanks

manisandro:
Thanks!

PR #279, opened 2023-12-01
Select format when printing layout

benoitblanc:
Hi !

This PR allows user to print layouts with other formats than PDF by selecting PDF, PNG, JPEG or SVG. By default, PDF is selected.

Please let me know if you want to set option configs for available formats and default format. 

Thanks


manisandro:
Thanks! A config option to restrict the available formats would make sense IMO.

PR #281, opened 2023-12-14
Use ctrl key to deselect a feature

cazitouni:
Hello, 

I made this PR to add the capability for the user to deselect an already selected feature visually using the ctrl key.
Before the ctrl key let the user select multiple feature but there was no way to deselect a feature visually, it was only doable using the minus button in the feature form.
I've tested localy using the demo app and it should work but let me know if the code isn't good enough or if my solution isnt the best. 

Thanks in advance for the review, 
Regards, Clément.


manisandro:
LGTM, thanks!

tpo:
This will not work on mobile phones though, since they do not have a CTRL key

manisandro:
Well no harm done, since also selection using ctrl will not work on mobiles.

PR #282, opened 2024-01-03
first steps to specify side parameter in plugins with resizeable windows

danceb:
In the end this should solve https://github.com/qgis/qwc2-demo-app/issues/510.

I've implemented the required parameters within the respective plugins and updated the resizeableWindow component, so the initial position of the window is different in relation to the docking side.

The initial position can now be set in relation to each side, where positive position values are related to top (InitialY) and left (InitialX) and negativ values to bottom (InitialY) and right (InitialX).

One thing is missing, where I could need some help/hints for implementation. Until now there is a fixed value of `100` to compensate the height of top bar and bottom bar. It would be very nice to get the exact height of these and calculate the height and position of the window in relation to this.
So everytime there is
```
- 100
```
It should be something like
```
- (topBarHeight + bottombarHeight)
```
But how can I get these? I've tried to get something using the _effect hook_ with `useState` and `useEffect`, but didn't come to working solution. Maybe you have an approach for this? Thanks in advance!

### DIFF
--- a/components/ResizeableWindow.jsx
+++ b/components/ResizeableWindow.jsx
@@ -79,14 +79,14 @@ class ResizeableWindow extends React.Component {
         this.dragShield = null;
         this.titlebar = null;
         this.id = uuidv1();
-        const height = Math.min(props.initialHeight, window.innerHeight - 100);
         const width = Math.min(props.initialWidth, window.innerWidth);
+        const height = Math.min(props.initialHeight, window.innerHeight - 100);
         if (WINDOW_GEOMETRIES[props.title]) {
             this.state.geometry = WINDOW_GEOMETRIES[props.title];
         } else {
             this.state.geometry = {
                 x: props.initialX !== null ? this.computeInitialX(props.initialX) : Math.max(0, Math.round(0.5 * (window.innerWidth - width))),
-                y: props.initialY !== null ? this.computeInitialY(props.initialY) : Math.max(0, Math.round(0.5 * (window.innerHeight - height))),
+                y: props.initialY !== null ? this.computeInitialY(props.initialY) : Math.max(0, Math.round(0.5 * (window.innerHeight - height - 100))),
                 width: width,
                 height: height,
                 docked: false
@@ -102,10 +102,18 @@ class ResizeableWindow extends React.Component {
         }
     }
     computeInitialX = (x) => {
-        return x >= 0 ? x : window.innerWidth - Math.abs(x);
+        if (this.props.dockable == "right") {
+            return x > 0 ? x : window.innerWidth - this.props.initialWidth - Math.abs(x);
+        } else {
+            return x >= 0 ? x : window.innerWidth - this.props.initialWidth - Math.abs(x);
+        }
     };
     computeInitialY = (y) => {
-        return y >= 0 ? y : window.innerHeight - Math.abs(y);
+        if (this.props.dockable == "bottom") {
+            return y > 0 ? y : window.innerHeight - this.props.initialHeight - Math.abs(y) - 100;
+        } else {
+            return y >= 0 ? y : window.innerHeight - this.props.initialHeight - Math.abs(y) - 100;
+        }
     };
     componentDidMount() {
         this.props.registerWindow(this.id);
--- a/plugins/MapLegend.jsx
+++ b/plugins/MapLegend.jsx
@@ -97,10 +97,10 @@ class MapLegend extends React.Component {
         ];
 
         return (
-            <ResizeableWindow extraControls={extraControls} icon="list-alt"
+            <ResizeableWindow dockable={this.props.geometry.side} extraControls={extraControls} icon="list-alt"
                 initialHeight={this.props.geometry.initialHeight} initialWidth={this.props.geometry.initialWidth}
                 initialX={this.props.geometry.initialX} initialY={this.props.geometry.initialY}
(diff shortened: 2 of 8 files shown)


plugins/MapLegend.jsx
```diff
@@ -95,9 +97,12 @@ class MapLegend extends React.Component {
         ];
 
         return (
-            <ResizeableWindow extraControls={extraControls} icon="list-alt" initialHeight={this.props.geometry.initialHeight} initialWidth={this.props.geometry.initialWidth}
+            <ResizeableWindow extraControls={extraControls} icon="list-alt"
+                initialHeight={this.props.geometry.initialHeight} initialWidth={this.props.geometry.initialWidth}
                 initialX={this.props.geometry.initialX} initialY={this.props.geometry.initialY}
-                initiallyDocked={this.props.geometry.initiallyDocked} onClose={this.onClose} title={LocaleUtils.trmsg("maplegend.windowtitle")} >
+                initiallyDocked={this.props.geometry.initiallyDocked} dockable={this.props.geometry.side}
```
manisandro, line 103: Please keep props sorted or else eslint will complain

danceb: Ah do you mean, they have to be in alphabetical order?

danceb:
> The second option is probably the easier/cleaner. I.e. adding an `id` to the topbar/bottombar elements and obtaining the height with `document.getElementById('xxx').clientHeight`.

As the bars have `ids` already, I have tried this in a similar way like this:
```
topBar = document.getElementById('TopBar');
topBarHeight = getComputedStyle(topBar).getPropertyValue("height");
```
But `document.getElementById('TopBar')` does not work as the result is `null`. 
I think, at the moment I try to get the element the rendering of the page isn't done and so I don't get it? It works, If I put the code within
```
setTimeout(() => {
  const topBar = document.getElementById('TopBar');
}); 
```
But thats not the best solution here.  I will try to have a look a the the variables suggestion.

plugins/MapLegend.jsx
manisandro: Yes exactly

manisandro:
Right, if the window component is mounted before the topbar/bottombar, then you run into this issue. A non-invasive variable solution would be to set the height of the bars as CSS variables, say:

    :root {
        --topbar-height: '3em'
    }

and then query the value via

    getComputedStyle(document.documentElement).getPropertyValue('--topbar-height');

danceb:
I have reordered the props. 

As I don't getting the reading of the height values to work unfortunately, I will leave this for a potential other PR.

danceb:
If you are fine with the other changes, I think it's ok, to merge the state here. The `100` I am trying to overcome here was there before my changes already. So I think this could be an improvement with another pull request. Thanks!

manisandro:
Fine with me, thanks!

components/ResizeableWindow.jsx
```diff
@@ -102,10 +102,18 @@ class ResizeableWindow extends React.Component {
         }
     }
     computeInitialX = (x) => {
-        return x >= 0 ? x : window.innerWidth - Math.abs(x);
+        if (this.props.dockable == "right") {
+            return x > 0 ? x : window.innerWidth - this.props.initialWidth - Math.abs(x);
+        } else {
+            return x >= 0 ? x : window.innerWidth - this.props.initialWidth - Math.abs(x);
+        }
```
manisandro, line 109: @danceb Reading this change more carefully, I cannot understand why it is needed? InitialX and initialY are actually only used in the floating state, so modifiying the behaviour based on the dockable state looks wrong. Also, say with `dockable=right`and `initialX=0`, having the left window border at the right screen border is most likely not what you want, since it would just mean that the window is out of sight. Can you elaborate the rationale for this?

manisandro:
Regarding topbar and bottombar height: please test with https://github.com/qgis/qwc2/commit/dc0607b711b30ddd0f483309bd809bcebc9d5c9c

components/ResizeableWindow.jsx
danceb: I thought, that it is more obvious, if the docking side is the one where the initial values are related to. So if you specify `dockable=right` and `initialX=0` without that change, it would place the floating window on the left hand side. Which is uncommon in my opinion as clicking the docking button will then dock the window to the opposite side and furthermore there is no chance to place the window on the right hand side.
With `return x > 0 ? x : window.innerWidth - this.props.initialWidth - Math.abs(x);` you will have the right window border at the right screen border, not the left one, as the `initialWidth` is also considered here. Calculating the position without `initialWidth` as it was before led to a placement out of sight. 
Does this make sense to you?

manisandro: Right, I missed the `- this.props.initialWidth` part. I'd still prefer to not mix the floating positioning with the docking state, an elegant solution IMO would be to treat negative 0 as right and positive zero as left (resp bottom/top). Indeed it is possible to distinguish between positive and negative zero via `Object.is`. What do you think?

danceb: Ah perfect! Didn't know about this option. So using `Object.is` will lead to this:
```javascript
    computeInitialX = (x) => {
        return x > 0 || Object.is(x, 0) ? x : window.innerWidth - this.props.initialWidth - Math.abs(x);
    };
    computeInitialY = (y) => {
        return y > 0 || Object.is(y, 0) ? y : window.innerHeight - this.props.initialHeight - Math.abs(y) - 100;
    };
```

PR #283, opened 2024-01-04
Store bottombar height in state

stamogis:
I needed the height of the bottombar for a custom plugin. Similar to the topbar, I added the storing of the bottombar height in state. Maybe this is also useful for other users.

### DIFF
--- a/actions/map.js
+++ b/actions/map.js
@@ -21,6 +21,7 @@ export const ZOOM_TO_POINT = 'ZOOM_TO_POINT';
 export const CHANGE_ROTATION = 'CHANGE_ROTATION';
 export const TOGGLE_MAPTIPS = 'TOGGLE_MAPTIPS';
 export const SET_TOPBAR_HEIGHT = 'SET_TOPBAR_HEIGHT';
+export const SET_BOTTOMBAR_HEIGHT = 'SET_BOTTOMBAR_HEIGHT';
 export const SET_SNAPPING_CONFIG = 'SET_SNAPPING_CONFIG';
 
 export function changeMapView(center, zoom, bbox, size, mapStateSource, projection) {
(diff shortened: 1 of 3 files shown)


manisandro:
Thanks! @danceb I actually completely forgot about this, this could be exactly what you need for #282?

danceb:
@stamogis May you give me an example how to access the height of the bottombar from another plugin? I am sorry, but I have no clue how this works ...

stamogis:
Hi @danceb, here is a small example plugin:

```JSX
import React from 'react';
import PropTypes from 'prop-types';
import {connect} from 'react-redux';

class PluginExample extends React.Component {
    static propTypes = {
        map: PropTypes.object
    };
    render() {
        const bottombarHeight = this.props.map.bottombarHeight;

        return (
            <div>...</div>
        );
    }
}

export default connect((state) => ({
    map: state.map
}), {})(PluginExample);
```

danceb:
Thank you so much! I hope that it's ok to ask another question based on that:
How do you ensure in your plugin that the bottom bar is rendered already before request its height? I try to get the  heights of top and bottom bar for the ResizableWindow component and at the time of its rendering the bars aren't rendered yet, so I get the value `0` for their heights...

manisandro:
Is this related to plugin activated via startup task? If yes, the solution would be to delay activating the startup task until the topbar/bottobar components are mounted - I can look into that.

stamogis:
Sorry, I can't help here. My plugin needs to be started manually by the user via the menu and then the bars should normally already be rendered. I don't have any extra handling if the two bars aren't rendered yet.

danceb:
> Is this related to plugin activated via startup task? If yes, the solution would be to delay activating the startup task until the topbar/bottobar components are mounted - I can look into that.

Unfortunately no. As I am trying to implement this within the ResizableWindow component, this effects all plugins which use this.

PR #289, opened 2024-01-05
Add declutter option to MVT

cazitouni:
I would like to submit this PR in order to add a config option to enable declutter or not. 
Not using declutter with a tilegrid may result to cropped and stacked labels. 

Example without declutter : 
Labels got stacked on each other
image
Labels got cropped at certain scales 
image

Example with declutter : 
City name under roads do not load, as intended and labels are rendered properly 
image
Labels are not cropped anymore
image

Let me know if it's ok for you or not. A modification in the documentation will be needed as well if it's ok.
Thanks in advance, 
Kind regards, Clément.


manisandro:
LGTM, thanks!

PR #290, opened 2024-01-07
calculation of initial position of resizable windows updated

danceb:
As mentioned in https://github.com/qgis/qwc2/pull/282#discussion_r1443778225 this brings the suggested changes, where `-0` is used for initial position instead of relating to the docking state.
Respective documentation in plugins were updated, too.


manisandro:
Thanks!

PR #292, opened 2024-01-11
Add option for the indent size behavior in layer catalog

stamogis:
With many subgroups in the layer catalog, the level based indent size becomes relatively large. So I added an option for a variant with a constant indent size.


manisandro:
Thanks

PR #293, opened 2024-01-11
Add css class corresponding to map button position

stamogis:
For the possibility to modify certain map buttons with custom css (qwc2.css), I added another css class based on the map button position.


manisandro:
Would `map-button:nth-child(n)` achieve the same without the need for extra classes?

stamogis:
Unfortunately, this only seems to work correctly if nothing changes in the DOM. In the current demo, for example, the "Layers & Legend" button would be n = 20. If I open the measurement tool, it would be n = 21.

manisandro:
Right, `nth-of-type` should work though?

stamogis:
thanks, that seems to work.

PR #296, opened 2024-01-12
Add support for complex "iframedialog" strings as target to the TopBar/BottomBar

stamogis:
Add property `iframeOptions` for menuItems/toolbarItems and `termsUrlOptions`/`viewertitleUrlOptions`for BottomBar to define the ResizeableWindow options (dockable, printable, etc.)

### DIFF
--- a/components/AppMenu.jsx
+++ b/components/AppMenu.jsx
@@ -193,7 +193,7 @@ class AppMenu extends React.Component {
             this.toggleMenu();
         }
         if (item.url) {
-            this.props.openExternalUrl(item.url, item.target, LocaleUtils.tr("appmenu.items." + item.key), {icon: item.icon, ...(item.iframeOptions || {})});
+            this.props.openExternalUrl(item.url, item.target, LocaleUtils.tr("appmenu.items." + item.key), item.icon);
         } else {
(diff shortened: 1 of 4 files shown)


manisandro:
You should already be able to specify the options in the `target` string as

    :iframedialog:<dialogname>:<optkey>=<optval>:<optkey>=<optval>:...

see [1]. Or am I missing something?

[1] https://qwc-services.github.io/topics/FeatureInfo/#attribute-values-html-markup-hyperlinks-images

stamogis:
Interesting, I didn't know that yet. Unfortunately, it doesn't seem to work with menu items. Is the problem here possibly that exactly "`iframe`" is expected as the target? 

https://github.com/qgis/qwc2/blob/970eecaa0ef7509d9fc114061e19e2935082c122/plugins/TopBar.jsx#L166

manisandro:
Ah right missed that. So yes, either the change you are proposing. or changing the TopBar/BottomBar to also accept a complex "iframedialog" string.

stamogis:
I reverted the previous changes and added additional support for the complex "iframedialog" strings instead.

manisandro:
Thanks!

PR #298, opened 2024-01-12
Reset current timestamp when theme has changed

benoitblanc:
Hi,

I have caught a bug with TimeManager, application crashes with following steps (you can reproduce on http://qwc2.sourcepole.ch) :

- load application with a theme with no time dimension ("Demo" theme)
- open time manager menu and keep it opened
- swith to a theme with time dimension ("Time manager demo" theme)
- application crashes

This PR fixes the problem. We need to set `currentTimestamp` to `null` if theme has changed.

Thanks


manisandro:
Thanks!

PR #299, opened 2024-01-16
Fix fallback filename in Print plugin

stamogis:
With `inlinePrintOutput` deactivated, the print plugin on chromium based web browsers always returns a `.pdf` file, no matter which print format is selected (jpeg, png, ...). Firefox doesn't have this problem.

There was the same problem with the MapExport plugin before (https://github.com/qgis/qwc2-demo-app/issues/489 / https://github.com/qgis/qwc2/commit/c8a7b3f9e60279b6eba24fc2217d86827d9d0b31). I added the solution from that issue to the print plugin as well.


manisandro:
Thanks!

PR #300, opened 2024-01-18
Do not automatically submit search for single parameter if text is manually typed 

cazitouni:
Hello @manisandro, 

Fist of all thanks for the quick fix about [this](https://github.com/qgis/qwc2-demo-app/issues/521).
But I would like to submit this PR to change it a bit if possible. 

For me, it seems a little "aggresive" to have the search automatically submited when user type text in the bar if there is only one result. 
People tend to make mistakes etc and you may want to verify your search result before letting the map moving you somewhere. 

But still it would be great to have the autosearch with the st= url parameters. 
So I've made a new state to determine if the search is triggered from st= or not. 

I don't know if my solution is optimal so let me know if it needs modification. 

Thanks in advance for your answer, 
Regards, Clément.

### DIFF
--- a/components/SearchBox.jsx
+++ b/components/SearchBox.jsx
@@ -88,7 +88,8 @@ class SearchBox extends React.Component {
         activeLayerInfo: null,
         filterOptionsVisible: false,
         selectedProvider: "",
-        selectedFilterRegion: null
+        selectedFilterRegion: null,
+        comesFromUrl: undefined
     };
     constructor(props) {
         super(props);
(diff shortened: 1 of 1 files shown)


manisandro:
I see your concern. However careful that with this approach the state will apply also to all subsequent searches if `st` was set. I've committed [1] to achieve the desired behaviour.

[1] https://github.com/qgis/qwc2/commit/8d858f80a38f1a0c8432b5d81686232b1a67785d

PR #301, opened 2024-01-19
Open native Identify tool from another Identify plugin

smaspons:
Hi!

A bit of context: I have a plugin that works similar to the Identify, it lets the user click the map and it returns some information on that point. If it doesn't return any results, it uses the qwc2 Identify to see if it might return something.

The thing is, if I open the Identify tool via another "identify" tool, when the results window is closed the tool must be closed, so the prop `exitTaskOnResultsClose` must be `true`. I set it to `true` in the action and I tried to still get the value defined in the `config.json`, but the [code](https://github.com/bgeo-gis/qwc2/blob/e8780643ded3931b7217eea333661d3d73ef469e/plugins/Identify.jsx#L441) is not pretty at all, I'm sure it can be handled much better, but I can't think of anything better...

There's also another problem. When the action sets the `exitTaskOnResultsClose` to `true`, it remains `true`, it doesn't go back to the value in the `config.json` when closing the window. Should it? If so, how?

Thanks! And please, let me know if there's a better solution.

### DIFF
--- a/plugins/Identify.jsx
+++ b/plugins/Identify.jsx
@@ -443,5 +443,5 @@ export default connect(selector, {
     changeSelectionState: changeSelectionState,
     removeMarker: removeMarker,
     removeLayer: removeLayer,
-    setCurrentTask: setCurrentTask,
+    setCurrentTask: setCurrentTask
 })(Identify);
--- a/plugins/Identify.jsx
+++ b/plugins/Identify.jsx
@@ -23,6 +23,7 @@ import TaskBar from '../components/TaskBar';
 import IdentifyUtils from '../utils/IdentifyUtils';
 import LocaleUtils from '../utils/LocaleUtils';
 import VectorLayerUtils from '../utils/VectorLayerUtils';
+import ConfigUtils from '../utils/ConfigUtils';
 
 import './style/Identify.css';
 
@@ -436,7 +437,8 @@ const selector = (state) => ({
     layers: state.layers.flat,
     map: state.map,
     selection: state.selection,
-    forceOpen: state.identify.forceOpen
+    forceOpen: state.identify.forceOpen,
+    exitTaskOnResultsClose: state.localConfig.plugins.desktop.find(x => x.name === "Identify").cfg.exitTaskOnResultsClose || state.identify.exitTaskOnResultsClose
 });
 
 export default connect(selector, {
--- a/plugins/Identify.jsx
+++ b/plugins/Identify.jsx
@@ -16,14 +16,12 @@ import { point as turfpoint } from "@turf/helpers";
 import {LayerRole, addLayerFeatures, addMarker, removeMarker, removeLayer} from '../actions/layers';
 import {changeSelectionState} from '../actions/selection';
 import {setCurrentTask} from '../actions/task';
-import {setForceOpen} from '../actions/identify';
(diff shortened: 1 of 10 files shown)


smaspons:
Hmm, I don't fully understand what this implies, but our identify tool has its own QtDesignerForm and its own logic so I'm not sure that adding a way to specify a custom query function would work for us, however it might be interesting for someone.

The only thing we need is a way to open the native Identify tool from our plugin, and with the changes I made it's possible to do so. And AFAIK it doesn't affect the Identify tool in any way if you don't use the action, that's why I made this PR...

manisandro:
I haven't fully tried it, but explicitly triggering the `Identify` task and extending the `Identify` Plugin to take additional task data should work? I.e.

    setCurrentTask("Identify", null, null, {"pos": [x, y]});

and then pass `this.state.task.data` to the `Identify` component, which will contain the desired point.

I'm not sure I fully understand the `exitTaskOnResultsClose` issue, but perhaps you can also influence this via task data?

smaspons:
You're right, this is much cleaner. With these last changes it's working for us. 
Please let us know if there's something wrong. Thanks!

plugins/Identify.jsx
```diff
@@ -304,8 +307,8 @@ class Identify extends React.Component {
             return {identifyResults: identifyResults};
         });
     };
-    onShow = (mode) => {
-        this.setState({mode: mode || 'Point'});
+    onShow = (mode, data) => {
+        this.setState({mode: mode || 'Point', clickPoint: data.pos, exitTaskOnResultsClose: data.exitTaskOnResultsClose});
```
manisandro, line 311: I think it could work in an even cleaner way by:
```
 if (this.props.currentIdentifyTool === "Identify" || this.props.currentTask === "Identify") {
             if (this.state.mode === "Point") {
+               const clickPoint = this.queryPoint(prevProps);
-               this.identifyPoint(prevProps);
+               this.identifyPoint(clickPoint);
             } else if (this.state.mode === "Region") {
```

and the modify `identifyPoint` below
```
-    identifyPoint = (prevProps) => {
+    identifyPoint = (clickPoint) => {
```
and invoke `identifyPoint(data.pos)` directly here instead of storing the `clickPoint` in the state.

jblanchg:
This works just fine, please let us know if there's something wrong. Thanks!

plugins/Identify.jsx
```diff
@@ -304,8 +305,9 @@ class Identify extends React.Component {
             return {identifyResults: identifyResults};
         });
     };
-    onShow = (mode) => {
-        this.setState({mode: mode || 'Point'});
+    onShow = (mode, data) => {
+        this.setState({mode: mode || 'Point', exitTaskOnResultsClose: data.exitTaskOnResultsClose});
+        this.identifyPoint(data.pos);
```
manisandro, line 310: I'd encolose this within an `if (mode === "Point" && data.pos)` just for easier code readability, for the rest LGTM!

smaspons: Okay, I figured that it wasn't necessary since `this.identifyPoint` already checks if the `clickPoint` exists, but yeah, for easier code readability I get it.
Should be good now, thanks!

manisandro:
Thanks!

PR #302, opened 2024-01-26
Bug fix: manage if data is null on identify onShow

smaspons:
Hi @manisandro. There was a bug on the previous pull request (#301) that chashed the app if data was null. This should fix it.

Sorry and thanks!


manisandro:
Thanks!

PR #303, opened 2024-01-30
Add an icon in themeswitcher to choose user default theme

gwenandres:
Hello, 
In this PR I propose to add a star icon (new.svg icon) on themes in themeswitcher to visualize and change default theme (for user logged).

Screenshot 2024-01-30 at 12-30-11 QGIS Web Client 2

I add an option (like in settings plugin) in themeswitcher plugin config to get or not this icon. 

What do you think about that ? 

Have a good day

### DIFF
--- a/components/ThemeList.jsx
+++ b/components/ThemeList.jsx
@@ -30,6 +30,7 @@ class ThemeList extends React.Component {
         allowAddingOtherThemes: PropTypes.bool,
         changeTheme: PropTypes.func,
         collapsibleGroups: PropTypes.bool,
+        defaultUrlParams: PropTypes.string,
         dontPreserveSettingsOnSwitch: PropTypes.bool,
         filter: PropTypes.string,
         layers: PropTypes.array,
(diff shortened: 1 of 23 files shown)


components/ThemeList.jsx
```diff
@@ -85,6 +89,9 @@ class ThemeList extends React.Component {
         const addLayersTitle = LocaleUtils.tr("themeswitcher.addlayerstotheme");
         const addTitle = LocaleUtils.tr("themeswitcher.addtotheme");
         const openTabTitle = LocaleUtils.tr("themeswitcher.openintab");
+        const changeDefaultUrlTitle = LocaleUtils.tr("themeswitcher.changedefaulttheme");
+        const username = ConfigUtils.getConfigProp("username");
+
```
manisandro, line 94: Please sort props alphabetically (resp. configure eslint in your editor).

manisandro:
Thanks, looks like a nice improvement

components/ThemeList.jsx
gwenandres: Sorry for that !

manisandro:
Thanks!

PR #307, opened 2024-02-07
Add prop to manage highlight on all identify results

smaspons:
Hi!

I found that the Identify highlight on all features was a bit too intense, so I added this prop `highlightAllResults` that if it's set to `false` it will only highlight the selected feature/layer. 
By default it's `true`, so it should work exactly like it did before.

Let me know if I did something wrong.
Thanks!


manisandro:
LGTM, thanks!

PR #308, opened 2024-02-07
An other way to open menu

gwenandres:
Hello, 
with @sylvainbeo we propose a new way to open main menu. 

If you add `appMenuCompact` in `TopBar` cfg, main menu is open in compact mode. 
On hover you will have all details .

menuCompact07-02-2024 18-18

I hope we can propose this for everyone.

Have a good day

### DIFF
--- a/actions/windows.js
+++ b/actions/windows.js
@@ -79,13 +79,12 @@ export function raiseWindow(id) {
     };
 }
 
-export function setSplitScreen(windowId, side, size, menuMargins) {
+export function setSplitScreen(windowId, side, size) {
     return {
         type: SET_SPLIT_SCREEN,
         windowId: windowId,
         side: side,
-        size: size,
-        menuMargins: menuMargins
+        size: size
     };
 }
 
--- a/actions/windows.js
+++ b/actions/windows.js
@@ -17,6 +17,7 @@ export const CLOSE_ALL_WINDOWS = 'CLOSE_ALL_WINDOWS';
 export const REGISTER_WINDOW = 'REGISTER_WINDOW';
 export const UNREGISTER_WINDOW = 'UNREGISTER_WINDOW';
 export const RAISE_WINDOW = 'RAISE_WINDOW';
+export const SET_MENU_MARGIN = 'SET_MENU_MARGIN';
 export const SET_SPLIT_SCREEN = 'SET_SPLIT_SCREEN';
 
 export const NotificationType = {
@@ -78,11 +79,20 @@ export function raiseWindow(id) {
     };
 }
 
-export function setSplitScreen(windowId, side, size) {
+export function setSplitScreen(windowId, side, size, menuMargins) {
     return {
         type: SET_SPLIT_SCREEN,
         windowId: windowId,
         side: side,
-        size: size
+        size: size,
+        menuMargins: menuMargins
+    };
+}
+
+export function setMenuMargin(right, left) {
+    return {
+        type: SET_MENU_MARGIN,
+        right: right,
+        left: left
     };
 }
--- a/components/AppMenu.jsx
+++ b/components/AppMenu.jsx
@@ -13,7 +13,7 @@ import mousetrap from 'mousetrap';
 import {remove as removeDiacritics} from 'diacritics';
 import isEmpty from 'lodash.isempty';
 import classnames from 'classnames';
-import {setRightMargin} from '../actions/map';
+import {setMenuMargin} from '../actions/windows';
 import {setCurrentTask} from '../actions/task';
 import InputContainer from '../components/InputContainer';
 import LocaleUtils from '../utils/LocaleUtils';
@@ -38,7 +38,7 @@ class AppMenu extends React.Component {
         onMenuToggled: PropTypes.func,
         openExternalUrl: PropTypes.func,
         setCurrentTask: PropTypes.func,
-        setRightMargin: PropTypes.func,
+        setMenuMargin: PropTypes.func,
         showFilterField: PropTypes.bool,
         showOnStartup: PropTypes.bool
     };
@@ -312,7 +312,7 @@ class AppMenu extends React.Component {
     storeRigthMargin = (el) => {
         if (this.props.menuCompact && el?.clientWidth > 0) {
             const rightmargin = el.clientWidth - MiscUtils.convertEmToPx(11.5);
-            this.props.setRightMargin(rightmargin);
+            this.props.setMenuMargin(rightmargin, 0);
         }
     };
     itemAllowed = (item) => {
@@ -337,5 +337,5 @@ export default connect((state) => ({
     currentTheme: state.theme.current || {}
 }), {
     setCurrentTask: setCurrentTask,
-    setRightMargin: setRightMargin
+    setMenuMargin: setMenuMargin
 })(AppMenu);
--- a/components/AppMenu.jsx
+++ b/components/AppMenu.jsx
@@ -13,6 +13,7 @@ import mousetrap from 'mousetrap';
 import {remove as removeDiacritics} from 'diacritics';
 import isEmpty from 'lodash.isempty';
 import classnames from 'classnames';
+import {setRightMargin} from '../actions/map';
 import {setCurrentTask} from '../actions/task';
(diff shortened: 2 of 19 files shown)


manisandro:
Thanks, I like the idea! I believe it should be restricted to desktop though?

components/PluginsContainer.jsx
```diff
@@ -20,6 +20,10 @@ class PluginsContainer extends React.Component {
         pluginsAppConfig: PropTypes.object,
         pluginsConfig: PropTypes.object
     };
+    isAppMenuCompact = (pluginsConfig)=> {
+        const pluginTopBar = pluginsConfig.find((pluginConf) => pluginConf.name === "TopBar");
+        return pluginTopBar && pluginTopBar.cfg.appMenuCompact;
+    };
```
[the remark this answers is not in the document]
gwenandres: Hello @manisandro, I think it is a better way to add an setRightMargin action. 
Thanks for your advice. 

I'm sorry but I think I didn't manage to set splitscreen=true. How/where to do that ? Or wich behavior in the interface ?  

I only have right-docked windows. 

Screenshot 2024-02-20 at 10-24-13 QGIS Web Client 2

Have a good day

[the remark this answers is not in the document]
manisandro: Hi @gwenandres , the `ResizeableWindow` component has a `splitScreenWhenDocked` prop. But the actual plugin rendering the window needs to set it, if it is useful for the component (i.e. AttributeTable, HeightProfile set `splitScreenWhenDocked`). For testing, you can just set it for a plugin which renders a right-dockable window.

components/AppMenu.jsx
```diff
@@ -173,9 +176,13 @@ class AppMenu extends React.Component {
...
         } else {
-            document.removeEventListener('click', this.checkCloseMenu);
-            document.removeEventListener('keydown', this.onKeyPress, true);
-            document.removeEventListener('mousemove', this.onMouseMove, true);
+            if (this.props.keepMenuOpen) {
+                return;
+            } else  {
+                document.removeEventListener('click', this.checkCloseMenu);
+                document.removeEventListener('keydown', this.onKeyPress, true);
+                document.removeEventListener('mousemove', this.onMouseMove, true);
```
manisandro, line 184: I just debugged an issue related to these event listeners. If they remain active, they will break arrow-pan of the map and i.e. enter-submit of input fields. So it is undesirable to have these event listeners active unless there really is an open menu.

gwenandres: Ok I will remove these changes

actions/windows.js
```diff
@@ -78,11 +79,20 @@ export function raiseWindow(id) {
...
 
-export function setSplitScreen(windowId, side, size) {
+export function setSplitScreen(windowId, side, size, menuMargins) {
     return {
         type: SET_SPLIT_SCREEN,
         windowId: windowId,
         side: side,
-        size: size
+        size: size,
+        menuMargins: menuMargins
```
manisandro, line 88: I don't think there is any advantage in passing the `menuMargins` here, since it is identical to the state value. Just use `menuMargins` in the reducer. You can move the `const mapMargins = {...}` block in the reducer to a separate funtion which you call both as a result of `SET_SPLIT_SCREEN` and `SET_MENU_MARGIN`.

components/AppMenu.jsx
```diff
@@ -168,14 +171,14 @@ class AppMenu extends React.Component {
...
-        } else {
+        if (this.state.menuVisible || this.props.keepMenuOpen) {
             document.removeEventListener('click', this.checkCloseMenu);
             document.removeEventListener('keydown', this.onKeyPress, true);
             document.removeEventListener('mousemove', this.onMouseMove, true);
+        } else {
+            document.addEventListener('click', this.checkCloseMenu);
+            document.addEventListener('keydown', this.onKeyPress, true);
+            document.addEventListener('mousemove', this.onMouseMove, true);
         }
```
manisandro, line 182: Please enclose the entire block between 174 and 181 in
```
if (!this.props.keepMenuOpen) {
...
}
```

components/AppMenu.jsx
```diff
@@ -302,6 +305,12 @@ class AppMenu extends React.Component {
             mousetrap(el).bind(this.props.appMenuShortcut, this.toggleMenu);
         }
     };
+    storeRigthMargin = (el) => {
```
manisandro, line 308: Rigth -> Right

components/AppMenu.jsx
```diff
@@ -302,6 +305,12 @@ class AppMenu extends React.Component {
...
+        if (this.props.menuCompact && el?.clientWidth > 0) {
+            const rightmargin = el.clientWidth - MiscUtils.convertEmToPx(11.5);
```
manisandro, line 310: What is 11.5em?

components/ResizeableWindow.jsx
```diff
@@ -169,7 +170,10 @@ class ResizeableWindow extends React.Component {
             "resizeable-window-body-scrollable": this.props.scrollable,
             "resizeable-window-body-nonscrollable": !this.props.scrollable
         });
-        const style = {display: this.props.visible ? 'initial' : 'none'};
+        const style = {
+            display: this.props.visible ? 'initial' : 'none',
+            right: 'calc(0.25em + ' + this.props.menuMargins.right + 'px)'
```
manisandro, line 175: What is 0.25em? Shouldn't this be part of the menuMargin, if necessary? Else with a zero menuMargin, it will still have a 0.25em right offset.
You'll also want to honour the left margin, since you support it in the reducer state.

components/SideBar.jsx
```diff
@@ -80,6 +81,7 @@ class SideBar extends React.Component {
         const render = visible || this.state.render || this.props.renderWhenHidden;
         const style = {
             width: this.props.width,
+            right: visible ? 'calc(0.25em + ' + this.props.menuMargins.right + 'px)' : 0,
```
manisandro, line 84: Same question regarding 0.25em as above

reducers/map.js
```diff
@@ -27,6 +27,7 @@ const defaultState = {
     resolutions: [0],
     topbarHeight: 0,
     bottombarHeight: 0,
+    rightMargin: 0,
```
manisandro, line 30: This looks like a leftover

reducers/windows.js
```diff
@@ -106,6 +110,19 @@ export default function windows(state = defaultState, action) {
...
+        const menuMargins =  {
+            right: action.right,
+            left: action.left
+        };
+        const mapMargins =  {
+            right: action.right,
+            bottom: 0,
+            left: action.left,
+            top: 0
+        };
```
manisandro, line 123: This looks wrong - why reset the top and bottom map margin to zero?

components/AppMenu.jsx
gwenandres: Hello, 
11.5em  is the right margin used in CSS to hide a part of menu : 
https://github.com/qgis/qwc2/pull/308/files#diff-9af8fe3b349e7d4c7e12d7406c0dc4898315de424adb8a5410a783c774ac807bR60

components/ResizeableWindow.jsx
gwenandres: I add 0.25em as it is done for MapCopyright : 

https://github.com/qgis/qwc2/blob/master/plugins/MapCopyright.jsx#L57

I found it aesthetically pleasing when everything wasn't aligned (stuck) with the right edge; even without compact menu.

I hope to honor the left margin but I can't do this right now. Should I remove left margin from reducer state ?

reducers/windows.js
gwenandres: Yes, I correct that

actions/windows.js
gwenandres: I'm sorry I don't understand how to move the `const mapMargins = {...}` block to a separate function. mapMargin Values in  `SET_SPLIT_SCREEN` and `SET_MENU_MARGIN` are different...

manisandro:
I've merged this manually and added a bunch of fixes, please check out if the latest master works for you. Thanks!

PR #310, opened 2024-02-20
Update themesConfig.py

vvmruder:
Fix minor bug. The fixed condition was true if configItem contains attribute `default` with value `false` which is not correct.


vvmruder:
I need to confirm (please wait before merge)

vvmruder:
You can try it yourself:

```python
"default" in {"default": False}
"default" in {"default": True}
{"default": False}.get("default", False)
{"default": True}.get("default", False)
```

manisandro:
Thanks

PR #311, opened 2024-02-26
Add filter layer param

jblanchg:
Hi, we created a custom plugin to filter layers and modified the CHANGE_LAYER_PROPERTY action to accept the 'FILTER' parameter and not only layers, opacities and styles.

Please let us know if there is an easier way to do it, thanks!


manisandro:
Here is a sample filter:
```
{"animals":[[["age",">","2"],"and",["age","<","99"]]]}
```

PR #316, opened 2024-02-28
Zoom to bookmark extent

benoitblanc:
Hi,

This PR allows user to only zoom to a bookmark extent without changing theme or load/open the bookmark in the current window or new tab.

Thanks for the review

### DIFF
--- a/plugins/Bookmark.jsx
+++ b/plugins/Bookmark.jsx
@@ -13,7 +13,7 @@ import classnames from 'classnames';
 import isEmpty from 'lodash.isempty';
 import PropTypes from 'prop-types';
 
-import { zoomToPoint } from '../actions/map';
+import { zoomToExtent, zoomToPoint } from '../actions/map';
 import Icon from '../components/Icon';
 import SideBar from '../components/SideBar';
 import Spinner from '../components/Spinner';
@@ -39,6 +39,7 @@ class Bookmark extends React.Component {
         side: PropTypes.string,
         state: PropTypes.object,
         task: PropTypes.string,
+        zoomToExtent: PropTypes.func,
         zoomToPoint: PropTypes.func
     };
     static defaultProps = {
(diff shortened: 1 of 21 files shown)


plugins/Bookmark.jsx
```diff
@@ -18,7 +18,9 @@ import Spinner from '../components/Spinner';
 import {createBookmark, getUserBookmarks, removeBookmark, updateBookmark} from '../utils/PermaLinkUtils';
 import './style/Bookmark.css';
 import isEmpty from 'lodash.isempty';
-
+import { zoomToPoint } from '../actions/map';
+import { resolveBookmark } from '../utils/PermaLinkUtils';
+import MapUtils from '../utils/MapUtils';
```
manisandro, line 24: Please keep imports sorted (first absolute, then relative, and lastly the style import)

plugins/Bookmark.jsx
```diff
@@ -118,6 +127,14 @@ class Bookmark extends React.Component {
             location.href = url;
         }
     };
+    zoomToBookmarkExtent = (bookmarkkey) => {
+        resolveBookmark(bookmarkkey, (params) => {
+            const scale = parseFloat(params.s);
+            const zoom = MapUtils.computeZoom(this.props.mapScales, scale);
+            const center = params.c.split(/[;,]/g).map(x => parseFloat(x));
```
manisandro, line 134: Depending on `urlPositionFormat` in `config.json`, the url parameters may contain `c` and `s`, or `e` for extent. You might also want to handle the `e` case here.

plugins/Bookmark.jsx
benoitblanc: Do we have some eslint rule for that ?

manisandro: Currently not, I suppose we could enable https://eslint.org/docs/latest/rules/sort-imports

benoitblanc: Ok I was looking to the same link. I can take the task ;-)

plugins/Bookmark.jsx
```diff
@@ -121,6 +130,19 @@ class Bookmark extends React.Component {
...
+    zoomToBookmarkExtent = (bookmarkkey) => {
+        resolveBookmark(bookmarkkey, (params) => {
+            if ('c' in params && 's' in params) {
+                const scale = parseFloat(params.s);
+                const zoom = MapUtils.computeZoom(this.props.mapScales, scale);
+                const center = params.c.split(/[;,]/g).map(x => parseFloat(x));
+                this.props.zoomToPoint(center, zoom, this.props.mapCrs);
+            } else if ('e' in params) {
+                const bounds = (params.e).split(',').map(n => parseFloat(n));
+                this.props.zoomToExtent(bounds, this.props.mapCrs);
```
manisandro, line 142: Sorry I forgot earlier, there is also the `urlPositionCrs` config option, which will add a `crs` URL parameter if `urlPositionCrs` is different than `mapCrs`. Something like `params.crs ?? this.props.mapCrs` should do.

benoitblanc: Thanks, I was not aware of these options

manisandro:
Thanks!

PR #319, opened 2024-03-01
Enable zoom by box selection

benoitblanc:
Hi,

This PR introduces a new `enableZoomByBoxSelection` option for `Zoom(In|Out)Plugin` to allow user to zoom by box selection. I needed to update `MapSelection` to make it work, I don't think I breaks something for other plugins that use this component but I'm not sure...

I tried to reproduce behavior for Zoom buttons in QGIS Desktop.

Here is a screencast with `enableZoomByBoxSelection: false` for `ZoomInPlugin` (actual behavior) and `enableZoomByBoxSelection: true` for `ZoomOutPlugin`:

QWCZoomByBox

Thanks for the review !

### DIFF
--- a/components/MapSelection.jsx
+++ b/components/MapSelection.jsx
@@ -21,7 +21,6 @@ class MapSelection extends React.Component {
     static propTypes = {
         /* Whether the selection tool is active */
         active: PropTypes.bool,
-        currentTask: PropTypes.string,
         /* Optional, a css-cursor to use when drawing */
         cursor: PropTypes.string,
         /* The selection geometry type (Point, LineString, Polygon, Circle, DragBox, Box) */
@@ -209,7 +208,6 @@ class MapSelection extends React.Component {
 }
 
 export default connect((state) => ({
-    currentTask: state.task.id,
     projection: state.map.projection
 }), {
 })(MapSelection);
--- a/components/MapSelection.jsx
+++ b/components/MapSelection.jsx
@@ -21,6 +21,7 @@ class MapSelection extends React.Component {
     static propTypes = {
         /* Whether the selection tool is active */
         active: PropTypes.bool,
+        currentTask: PropTypes.string,
         /* Optional, a css-cursor to use when drawing */
         cursor: PropTypes.string,
         /* The selection geometry type (Point, LineString, Polygon, Circle, DragBox, Box) */
@@ -103,6 +104,7 @@ class MapSelection extends React.Component {
     }
     componentWillUnmount() {
         this.map.removeLayer(this.selectionLayer);
+        this.removeDrawInteraction();
     }
     addDrawInteraction = () => {
         // cleanup old interaction
@@ -112,7 +114,9 @@ class MapSelection extends React.Component {
         if (this.props.geomType === "DragBox") {
             this.drawInteraction = new ol.interaction.DragBox({
                 className: 'selection-drag-box',
-                condition: ol.events.condition.shiftKeyOnly
+                condition: () => {
+                    return ["ZoomIn", "ZoomOut"].includes(this.props.currentTask) || ol.events.condition.shiftKeyOnly;
+                }
             });
 
             this.drawInteraction.on('boxend', () => {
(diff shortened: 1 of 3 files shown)


components/MapSelection.jsx
```diff
@@ -112,7 +114,9 @@ class MapSelection extends React.Component {
         if (this.props.geomType === "DragBox") {
             this.drawInteraction = new ol.interaction.DragBox({
                 className: 'selection-drag-box',
-                condition: ol.events.condition.shiftKeyOnly
+                condition: () => {
+                    return ["ZoomIn", "ZoomOut"].includes(this.props.currentTask) || ol.events.condition.shiftKeyOnly;
+                }
```
manisandro, line 119: I think you can generally simplify the condition to `return true`, `shiftKeyOnly` looks like leftover code, and I don't think it makes any sense here

plugins/ZoomButtons.jsx
```diff
@@ -26,17 +29,58 @@ import './style/Buttons.css';
...
+            if (this.props.direction > 0) {
+                this.setState({disabled: this.props.currentZoom >= this.props.maxZoom, task: "ZoomIn"});
+            } else if (this.props.direction < 0) {
+                this.setState({disabled: this.props.currentZoom <= 0, task: "ZoomOut"});
+            }
+        }
+        if (this.props.currentTask !== null && this.props.currentTask === this.state.task && this.state.disabled) {
+            this.props.setCurrentTask(null);
+        }
+        if (this.props.currentTask === this.state.task && prevProps !== this.props) {
```
manisandro, line 71: If you write `this.props.click && this.props.click !== prevProps.click` here and directly use  `this.props.click.coordinate` below you can simplify the code

manisandro:
A general remark: Does this conflict/overlap with the already implemented shift+drag zoomin?

components/MapSelection.jsx
benoitblanc: I thought it was used by the shift+drag zoom already implemented but nope. I have removed all the condition

plugins/ZoomButtons.jsx
benoitblanc: Resolved

components/MapSelection.jsx
```diff
@@ -21,6 +21,7 @@ class MapSelection extends React.Component {
     static propTypes = {
         /* Whether the selection tool is active */
         active: PropTypes.bool,
+        currentTask: PropTypes.string,
```
manisandro, line 24: This can also be removed (as well as below)

plugins/ZoomButtons.jsx
```diff
@@ -26,18 +29,53 @@ import './style/Buttons.css';
...
+            if (this.props.direction > 0) {
+                this.setState({disabled: this.props.currentZoom >= this.props.maxZoom, task: "ZoomIn"});
+            } else if (this.props.direction < 0) {
+                this.setState({disabled: this.props.currentZoom <= 0, task: "ZoomOut"});
+            }
+        }
+        if (this.props.currentTask !== null && this.props.currentTask === this.state.task && this.state.disabled) {
+            this.props.setCurrentTask(null);
+        }
+        if (this.props.currentTask === this.state.task && prevProps !== this.props && this.props.click !== prevProps.click) {
```
manisandro, line 71: > this.props.currentTask === this.state.task && this.props.click && this.props.click !== prevProps.click

benoitblanc: Thanks :facepalm:

benoitblanc:
> I'd propose these additional changes [1], what do you think?
> 
> [1] [db228da](https://github.com/qgis/qwc2/commit/db228daa1386370ad6885b437467b329c94b46d6)

Yes I agree with all of your changes, thanks.

For `don't automatically exit task when zoom limit reached`, I added it at the end of my PR because I wanted to exit the task to not have a `zoom-in|out` cursor for the user when he could not zoom because he has reached the limit, but I'm ok to keep this as current task. So user could still draw drag-box but zoom should not have effect I guess

manisandro:
>  So user could still draw drag-box but zoom should not have effect I guess

It will just re-center the map without changing the zoom level, which I think is not a bad behaviour.

manisandro:
Manually merged, thanks!

PR #321, opened 2024-04-09
Do not change protocol of urls in catalog.json

copierrj:
While investigating a CORS error that occurred when adding a layer using the LayerCatalog we discovered that the protocol part of urls in catalog.json are effectively ignored. 

When QWC is running at http://localhost:8081 (i.e. a typical development setup) and the following catalog.json is used as input for the LayerCatalog,

```json
{
  "catalog": [
      {
          "title": "Natuurnetwerk Nederland (NNN) - Provincies (INSPIRE geharmoniseerd) WMS",
          "resource": "wms:https://service.pdok.nl/provincies/natuurnetwerk-nederland/wms/v1_0#PS.ProtectedSite"
      }
    ]
}
```

then the configured service is accessed over HTTP and not (as configured) over HTTPS. The resulting redirect (302) from HTTP to HTTPS is causing CORS issues. 

This PR changes the behavior of the LayerCatalog to always just use the configured protocol.

### DIFF
--- a/utils/ServiceLayerUtils.js
+++ b/utils/ServiceLayerUtils.js
@@ -355,7 +355,9 @@ const ServiceLayerUtils = {
             };
             delete urlParts.search;
         }
-        urlParts.protocol = location.protocol;
+        if (urlParts.protocol === "http" && location.protocol === "https") {
+            urlParts.protocol = "https";
+        }
         const requestUrl = url.format(urlParts);
         return new Promise((resolve, reject) => {
             axios.get(requestUrl).then(response => {


manisandro:
So the reason for overriding the protocol is to avoid mixed content errors when a QWC2 instance running on HTTPS attempts to load a WMS via HTTP. What should fix this issue while still preventing mixed content issues is to:

- Either only promote HTTP schemes to HTTPS, and never demote HTTPS to HTTP
- Or try loading first with the matched SCHEME, and if that fails with the original scheme

copierrj:
Never demoting HTTPS to HTTP works for me. I'll change this PR accordingly.

manisandro:
Thanks!

PR #322, opened 2024-04-16
Add configurable timeout to process notifications

jblanchg:
Following PR #274. 

Configurable process notification's timeout. By default set to 'True', leaving the timeout times for clearing the process as they are, a number to set the desired time, and 'false' to not clear the process.


manisandro:
Thanks!

PR #323, opened 2024-04-26
Import zipped shapefile

benoitblanc:
Hi,

This PR aims to add SHP import in QWC. I have used [@loaders.gl](https://loaders.gl/) library to do so. User has to import ZIP file containing SHP file and sidecar files (shx, dbf, cpg, prj). There could be multiple SHP files in the ZIP.

I have also fixed a missing attribute which was not parsed using fast-xml-parser.

Thanks

### DIFF
--- a/components/ImportLayer.jsx
+++ b/components/ImportLayer.jsx
@@ -9,6 +9,9 @@
 import React from 'react';
 import {connect} from 'react-redux';
 
+import {_BrowserFileSystem as BrowserFileSystem, load} from '@loaders.gl/core';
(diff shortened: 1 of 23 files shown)


manisandro:
Nice! Maybe it would be a good idea to add an info text or an info icon with a tooltip regarding supported formats, which are otherwise hardly discoverable?

benoitblanc:
@manisandro Yes, it would be good to see all the supported formats. Do you think we could change translation of `fileselector.placeholder` or would it be better to have a new text/icon ?

manisandro:
Thanks!

PR #324, opened 2024-04-30
Adjust BackgroundSwitcher items also according to splitscreen configuration

stamogis:
The items for changing the background layers are hidden by the open height profile window. Similarly to https://github.com/qgis/qwc2/commit/a196ae6e0bafe6564276667676fcc0d3cacc24f2 also adjust the items according to the splitscreen configuration.


manisandro:
Thanks!

PR #325, opened 2024-06-11
Add ESLint to dev dependencies and fix linting violations

BenediktSeidlSWM:
I tried to apply linting adjustments or disable linting in accordance with the standards of this repository. Note that the directory `utils/expr_grammar/` is now completely excluded from linting since it only consists generated code only.


manisandro:
Thanks!

PR #327, opened 2024-07-03
Save Bookmarks on Enter

NinaRoeckeleinSWM:
UX Improvement:
Save a bookmark by hitting "Enter" after entering a name. Reduces necessary mouse clicks.
Assures that the description is not empty.


manisandro:
LGTM thanks

PR #329, opened 2024-07-12
Add translation for redlining draw layer

cazitouni:
Hello @manisandro,

I would like to submit this PR because I received feedback from many users who don't understand why the default redlining layer is named like this, even in other languages. It can be confusing for people who don't understand basic English.

So, this PR aims to use the translation tools to name the layer according to the user's browser language. I hope the code is correct, but please let me know if it isn't.

Thanks in advance,
Regards, Clément.

### DIFF
--- a/plugins/map/RedliningSupport.jsx
+++ b/plugins/map/RedliningSupport.jsx
@@ -20,9 +20,10 @@ import NumericInputWindow from '../../components/NumericInputWindow';
 import {OlLayerAdded, OlLayerUpdated} from '../../components/map/OlLayer';
 import displayCrsSelector from '../../selectors/displaycrs';
 import FeatureStyles from '../../utils/FeatureStyles';
+import LocaleUtils from '../../utils/LocaleUtils';
 import MapUtils from '../../utils/MapUtils';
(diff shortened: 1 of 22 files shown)


plugins/map/RedliningSupport.jsx
```diff
@@ -574,7 +575,7 @@ class RedliningSupport extends React.Component {
         feature.crs = this.props.mapCrs;
         const layer = {
             id: this.props.redlining.layer,
-            title: this.props.redlining.layerTitle,
+            title: LocaleUtils.tr(this.props.redlining.layerTitle),
```
manisandro, line 578: This line here is incorrect, layerTitle is already translated (see line 81 in Redlining.jsx).

cazitouni: I belived too, but If I do not use it here It goes back to it's untranslated state as soon as you add a point in the layer.

manisandro:
Should be done in https://github.com/qgis/qwc2/commit/1c591531c52577a9d6b4b96ca885c304d7d3bce1 and https://github.com/qgis/qwc2/commit/52d942f1b7555e06c1de2c957c93833f51bb3a6c

PR #330, opened 2024-07-22
Select layers when clicking on title

BenediktSeidlSWM:
Small UX improvement to allow selecting a layer group by clicking on its title instead of the checkbox


manisandro:
Why just group layers?

BenediktSeidlSWM:
Good point, I added the option for individual layers as well 👍

manisandro:
Thanks!

PR #332, opened 2024-09-03
Customizable providers label through theme config.

NullSeile:
Allow the configuration of a provider in the theme to override the `label` or `labelmsgid`.


manisandro:
LGTM, thanks

PR #333, opened 2024-09-17
Using theme id instead of name as filename for print and map export

danceb:
AS mentioned in [1] this prevents, that `pg_schemaname` is part of the filename, if the QGIS project is stored in database.

[1] https://github.com/qgis/qwc2-demo-app/issues/580#issuecomment-2352218604


manisandro:
Thanks

PR #335, opened 2024-09-26
Float precision rotation for print

cazitouni:
Hello @manisandro,

Some users of mine have requested more precision for the print rotation to better align buildings and other map elements with the template frame.

I've made a modification to address this. Let me know if it's okay for you or if you prefer to keep the precision as integers.

Kind regards,
Clément


manisandro:
Thanks

PR #336, opened 2024-10-04
Add an option to disable automatic zoom in routing plugin 

cazitouni:
Hello @manisandro,

I would like to submit this PR to add an option in the routing plugin to disable automatic zoom.

In some use cases, it could be more convenient for the user to keep the map from adjusting to the calculated road extent. For example, if they want to add another point farther away or if they have other elements on the map and do not want the scale to change.

I will update the documentation accordingly if this is approved.

Please feel free to share any feedback.

Kind regards,
Clément

### DIFF
--- a/plugins/Routing.jsx
+++ b/plugins/Routing.jsx
@@ -813,7 +813,7 @@ class Routing extends React.Component {
                 }));
                 this.props.addLayerFeatures(layer, features, true);
                 if (this.props.zoomAuto) {
-                    this.props.zoomToExtent(result.summary.bounds, "EPSG:4326", -0.5);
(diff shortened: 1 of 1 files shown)


plugins/Routing.jsx
```diff
@@ -60,6 +60,8 @@ class Routing extends React.Component {
             initiallyDocked: PropTypes.bool,
             side: PropTypes.string
         }),
+        /** Automatic zoom to the extent of the location. */
```
manisandro, line 63: What about "Automatically zoom to the extent of the route"?

cazitouni: Sure, no problem.

cazitouni:
Should be good now

manisandro:
Thanks!

PR #338, opened 2024-10-14
Restore clickable buttons on touch devices

BenediktSeidlSWM:
The `cancel` attribute got removed in a previous commit, removing the ability to close popup windows on mobile and touch devices.


manisandro:
Thanks!

PR #340, opened 2024-11-20
Rework print functionality

BenediktSeidlSWM:
This pull request includes a complete rework of the printing functionality of qwc2. Instead of an HTML overlay, the new PrintSelection component renders a new layer on the map. Inside this layer, the user can now interact with the selection box. It can be moved, scaled and rotated on the map without moving the map itself. This change provides a more intuitive and flexible way to print or export parts of the map.

Additionally, a new feature has been implemented to allow printing a series of selection boxes next to each other. This "print series" feature is disabled by default, but can be turned on with the `displayPrintSeries` flag on the Print plugin. When enabled, the user can select additional frames (optionally with an adjustable overlap). The user can then also select whether they want to obtain a PDF with several pages, multiple PDF documents with a single page or a ZIP archive containing multiple PDF documents.

**Additional changes**

A new option `fileNameTemplate` is introduced. This option allows us to set a custom file name for the generated documents. The default value is the template used right now.

The MapExport plugin has been updated to use the new PrintSelection component. It resembles the functionality of the old implementation with little updates on the styling.

The deprecated DxfExport and RasterExport plugins are removed. They do not seem to be completely functional any more and depend on the now obsolete PrintFrame component.

**Technical details**

The solution of setting a `modifyGeometry` property on the feature in combination with the ModifyInteraction provided by openlayers is inspired by [this example](https://openlayers.org/en/latest/examples/modify-scale-and-rotate.html).

The slightly tedious logic for rendering the background when the print series overlap is enabled is due to the fact that in openlayers, if two negative linear rings of a polygon overlap, the intersection is rendered as a positive part of the polygon.

**Screenshots**

image
image
image
image


tpo:
nice!!!

BenediktSeidlSWM:
Thanks for your positive feedback! Please let me know if you find any bugs or improvements. 😄

PR #342, opened 2024-12-13
Layer selection in identifyViewer.jsx

cazitouni:
Hello @manisandro,

I would like to submit this PR to address https://github.com/qgis/qwc2-demo-app/issues/593.

Please let me know if the code needs any modifications or if you have any questions about it.
I hope it'll work for you.

Kind regards,
Clément.

### DIFF
--- a/components/IdentifyViewer.jsx
+++ b/components/IdentifyViewer.jsx
@@ -197,7 +197,7 @@ class IdentifyViewer extends React.Component {
         selectedAggregatedReport: null,
         generatingReport: false,
         selectedLayer: 'all',
-        originalResultTree : {}
+        originalResultTree: {}
     };
     constructor(props) {
         super(props);
@@ -561,7 +561,7 @@ class IdentifyViewer extends React.Component {
             body = (
                 <div className="identify-flat-results-list">
                     <div className="identify-selectbox">
-                        <select name="identify-layer-select" id="identify-layer-select"
+                        <select className="identify-layer-select"
                             onChange={(e) => {const selectedLayer = e.target.value; this.setState({ selectedLayer });}}
                         >
                             <option value="all">{LocaleUtils.tr("identify.layerall")}</option>
--- a/components/IdentifyViewer.jsx
+++ b/components/IdentifyViewer.jsx
@@ -196,8 +196,7 @@ class IdentifyViewer extends React.Component {
         exportFormat: 'geojson',
         selectedAggregatedReport: null,
         generatingReport: false,
-        selectedLayer: 'all',
-        originalResultTree: {}
+        selectedLayer: ''
     };
     constructor(props) {
         super(props);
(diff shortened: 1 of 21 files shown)


components/IdentifyViewer.jsx
```diff
@@ -525,6 +560,21 @@ class IdentifyViewer extends React.Component {
         } else {
             body = (
                 <div className="identify-flat-results-list">
+                    <div className="identify-selectbox">
+                        <select className="identify-layer-select"
+                            onChange={(e) => {const selectedLayer = e.target.value; this.setState({ selectedLayer });}}
+                        >
+                            <option value="all">{LocaleUtils.tr("identify.layerall")}</option>
```
manisandro, line 567: There is a theoretical possibility of a name collision with a layer named `all`. I'd just use an empty string.

components/IdentifyViewer.jsx
```diff
@@ -298,16 +318,31 @@ class IdentifyViewer extends React.Component {
...
         });
     };
+    filterResult = () => {
+        let newResultTree;
+        if (this.state.selectedLayer === "all") {
+            newResultTree = this.state.originalResultTree;
+        } else {
+            newResultTree = { [this.state.selectedLayer]: this.state.originalResultTree[this.state.selectedLayer] };
+        }
+        this.setState({resultTree: newResultTree});
```
manisandro, line 344: You should not (directly or indirectly) use `this.state` in `setState`. Rather i.e.
```
this.setState((state) => {
     return {resultTree: state.originalResultTree};
});
```
Aside from this, I think it would be simpler to just filter
```
const contents = Object.keys(this.state.resultTree).map(layer => this.renderLayer(layer));
```
according to the selected layer in `render` instead of storing a separate result tree in the state.

components/IdentifyViewer.jsx
cazitouni: Thansk, I've fixed that

components/IdentifyViewer.jsx
cazitouni: Ok my solution was actually for the flast list, I wasn't really aware of that DisplaylyerTree. Then all I need is the possibility of showing all the result of the same layer maybe with a little button next to the delete one. 

I'll try something new then i'll let you know.

cazitouni:
I’ve taken a new approach to implementing this functionality based on your feedback and after reviewing the "Display Result Tree" option.

I still believe it would be a good idea to implement the filter on the flat list.

I’ve introduced a single new state called selectedLayer and used it to filter directly within the component's body.

Let me know if this approach is better or if further modifications are needed.

This change should not affect the "Result Tree" mode, only the flat list mode.

Also, would it be a good idea to create a clean commit tree and open a new PR for this?

Thanks again for your insights!

Best regards,
Clément

manisandro:
Thanks looks good!

manisandro:
Hmm actually I wonder, should the feature count not reflect the selected layer?

cazitouni:
> Hmm actually I wonder, should the feature count not reflect the selected layer?

Yes it should, and it was, I didn't notice during the refactoring. 

I'll fix that tommorow !

PR #343, opened 2024-12-18
Fix function call for point measuring

stamogis:
See moved function in commit 9a15944728640e51ecc616f1762e9679fb7df7d6


manisandro:
Thanks!

PR #347, opened 2025-01-29
Build filter from array

nperez-bgeo:
Hi, I wanted to set a WMS filter from `["muni_id", "IN", muniFilter]` where **muniFilter** is an array. With the the following WMS documentation in mind:

`Text strings need to be enclosed with quotes (single quotes for strings, double quotes for attributes) A space between each word / special character is mandatory. Allowed Keywords and special characters are ‘AND’,’OR’,’IN’,’=’,’<’,’>=’, ‘>’,’>=’,’!=*,’(‘,’)’. Semicolons in string expressions are not allowed`

And with the result filter I needed `"muni_id" IN ( 1 , 2 ) `,  I added the case to format the filter accordingly.

I suppose there is a better way to do it? Please let me know.

Thanks!

### DIFF
--- a/utils/LayerUtils.js
+++ b/utils/LayerUtils.js
@@ -224,6 +224,8 @@ const LayerUtils = {
                 return `"${expr[0]}" ${op} ${expr[2]}`;
             } else if (expr[2] === null) {
                 return `"${expr[0]}" ${op} NULL`;
+            } else if (Array.isArray(expr[2]) ){
+                return `"${expr[0]}" ${op} ( ${expr[2].join(' , ')} )`;
             } else {
                 return `"${expr[0]}" ${op} '${expr[2]}'`;
             }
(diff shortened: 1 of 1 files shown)


manisandro:
AFAICS this should work fine. Could you just fix the closing brace spacing please?
Thanks!

nperez-bgeo:
Hi, I tried within spaces but is not working. The space between special characters is necessary. I followed the next documentation:
`A space between each word / special character is mandatory.`

utils/LayerUtils.js
```diff
@@ -224,6 +224,8 @@ const LayerUtils = {
                 return `"${expr[0]}" ${op} ${expr[2]}`;
             } else if (expr[2] === null) {
                 return `"${expr[0]}" ${op} NULL`;
+            } else if (Array.isArray(expr[2]) ){
```
manisandro, line 227: I mean just the code style here

nperez-bgeo: Okay, sure!

manisandro: And also add the space between `)` and `{` please (do you have eslint configured in your editor?)

nperez-bgeo (commented):
fix: add space

utils/LayerUtils.js
nperez-bgeo: Sorry! I used the Github built-in editor as it was only a line of code. My bad.

manisandro:
Thanks!

PR #348, opened 2025-01-30
Panoramax plugin

cazitouni:
Hello @manisandro,

I am taking the liberty to submit this draft PR to get your opinion on the Panoramax plugin I've created.

I believe there could be a few improvements in the future, but it will be a good first implementation of Panoramax.

For instance, highlighting sequences, managing more tile styles, and navigating through dates/sequences within the image, etc.

On a more critical note, I still have no idea how to handle MVT tiles in a projection different from EPSG:3857.
See issue --> https://github.com/qgis/qwc2-demo-app/issues/600
Additionally, the poor MVT performance of OpenLayers (https://github.com/qgis/qwc2-demo-app/issues/503) could result in some lag when using MVT in an area with a lot of sequences at certain scales.
Maybe find a way or submit a modification to panoramax API to send a miror wms could be a solution ? 

Also, is there a reason the image style anchor was set to a fixed value of `[0.5,1]` instead of using `options.anchor` in `FeatureStyles.js`? I've modified it because I need that for the orientation symbol. I've seen no impact on the rest of the app so far, but I do not know all of it.

I've fixed a small bug in the Panoramax web viewer here --> https://gitlab.com/panoramax/clients/web-viewer/-/merge_requests/61
So, maybe waiting for the next release before merging anything would be a good idea.

Panoramax itself also has a few bugs that will need solving directly in the viewer repo. For example, in Chrome, it selects everything when you move your mouse out of the viewport of the sphere, and on some setups, the cursor is always a closed hand, which feels weird.

And finally, I didn't find any way to make the viewer work when you pop out the resizable windows of QWC. I do not know if there is any solution to that or if we should just disable this option for this plugin.

Sorry for the long explanation ! 
let me know what you think about that. 

Kind regards, 
Clément.


manisandro:
Hi @cazitouni , thanks for the PR!

Regarding MVT here is a working configuration of a background layer, maybe it helps?
```
      {
        "name": "lightbasemap_vt_swisstopo",
        "title": "Light Basemap",
        "type": "mvt",
        "url": "https://vectortiles.geo.admin.ch/tiles/ch.swisstopo.base.vt/v1.0.0/{z}/{x}/{y}.pbf",
        "style": "https://vectortiles.geo.admin.ch/styles/ch.swisstopo.lightbasemap.vt/style.json",
        "declutter": false,
        "thumbnail": "img/mapthumbs/swisstopo_lbm.png",
        "attribution": "© swisstopo",
        "attributionUrl": "https://www.geo.admin.ch/de/allgemeine-nutzungsbedingungen-bgdi",
        "projection": "EPSG:3857",
        "tileGridConfig": {
          "origin": [-20037508.342789244, 20037508.342789244],
          "resolutions": [
            78271.51696402048,
            39135.75848201024,
            19567.87924100512,
            9783.93962050256,
            4891.96981025128,
            2445.98490512564,
            1222.99245256282,
            611.49622628141,
            305.748113140705,
            152.8740565703525,
            76.43702828517625,
            38.21851414258813,
            19.109257071294063,
            9.554628535647032,
            4.777314267823516,
            2.388657133911758,
            1.194328566955879,
            0.5971642834779395,
            0.29858214173896974,
            0.14929107086948487,
            0.07464553543474244,
            0.03732276771737122,
            0.01866138385868561
          ]
        }
      }
```

> Also, is there a reason the image style anchor was set to a fixed value of [0.5,1] instead of using options.anchor in FeatureStyles.js? I've modified it because I need that for the orientation symbol. I've seen no impact on the rest of the app so far, but I do not know all of it.

Would indeed make sense to keep it configurable.

> And finally, I didn't find any way to make the viewer work when you pop out the resizable windows of QWC. I do not know if there is any solution to that or if we should just disable this option for this plugin.

I can also reproduce the issue with the Cyclomedia plugin, I'll have a look.

manisandro:
Re pop out window: with ` usePortal={false}` the Cyclomedia plugin mostly works (it reloads and forgets the currently loaded image, but thereafter you can use it normally)

cazitouni:
Re @manisandro ,

For the MVT, the method I use is actually good to calculate the resolutions for EPSG:3857. The problem is when the theme doesn't use EPSG:3857, the MVT is projected in the wrong place, which seems logical. QGIS Desktop handles this by reprojecting on the fly, but I do not know if it is possible to do that in OpenLayers. I was thinking maybe recalculating the resolutions for the theme's EPSG could help. But I do not know; maybe it's not a solvable problem in the current state of things.

For the pop-out, I've tried with usePortal set to false, and it now pops up the window with an empty DOM. With usePortal set to true, I get a good viewer with clickable buttons, etc., but I think it's an event listener issue because the viewer moves when I click on the popup and then move my mouse over the map. It's like the event listener for the viewer's mouse move is still on the map window but not the rest. I do not know if it's super clear.

I will try to look more at it this weekend if I have time.

Thanks for the feedback!

pka:
Hi @cazitouni, Panoramax is a great procect, hanks for your contribution!

AFAIK, Openlayers doesn't support reprojecting MVTs, right, @ahocevar?

ahocevar:
@pka

> AFAIK, Openlayers doesn't support reprojecting MVTs, right, @ahocevar?

Not yet at least :-)

manisandro:
> For the pop-out, I've tried with usePortal set to false, and it now pops up the window with an empty DOM. With usePortal set to true, I get a good viewer with clickable buttons, etc., but I think it's an event listener issue because the viewer moves when I click on the popup and then move my mouse over the map. It's like the event listener for the viewer's mouse move is still on the map window but not the rest. I do not know if it's super clear.

The event listener problem is a bit tricky, indeed if you do
```
window.addEventListener('xxx')
```
`window` will refer to the main browser window, not the popped out window. Instead, you need to add the listener to `ev.view`, see for instance [1].

This is relatively easy for self-written code, it is more problematic in thirdparty libraries which may need patching.

[1] https://github.com/qgis/qwc2/blob/master/components/SideBar.jsx#

cazitouni:
> What does this plugin do? Are there screenshots of it or a screencast or documentation of it somewhere?

Hello @tpo, 

The aim of this plugin is to implement support that allows users to consult images from a Panoramax instance in their QWC2 themes.

Panoramax is an open-source application that aims to centralize equirectangular (or other) images from all over the world. You can self-host an instance for yourself or your organization, or simply use a public one.
OSM and IGN instances are aggregated at https://api.panoramax.xyz/.

For now, it may be a bit premature since it's a draft, but it will, of course, be documented in the appropriate section if the plugin is completed.
Also, I will need to include some SVG assets in the demo app.

For usage, it's straightforward. The user clicks on the plugin, which loads the vector tiles with the sequences over the theme. Then, the user can select an image from any sequence, navigate through it via the viewer, or simply select another image on the map. When finished, the user closes the window, and the vector tiles unload.

I made a small clip 
[Enregistrement d'écran_20250130_205514.webm](https://github.com/user-attachments/assets/39588d96-f4e4-4392-ba94-ddbbc1f3f1e0)

Let me know if you need anything else. Thank for your feedback.

cazitouni:
> > For the pop-out, I've tried with usePortal set to false, and it now pops up the window with an empty DOM. With usePortal set to true, I get a good viewer with clickable buttons, etc., but I think it's an event listener issue because the viewer moves when I click on the popup and then move my mouse over the map. It's like the event listener for the viewer's mouse move is still on the map window but not the rest. I do not know if it's super clear.
> 
> The event listener problem is a bit tricky, indeed if you do
> 
> ```
> window.addEventListener('xxx')
> ```
> 
> `window` will refer to the main browser window, not the popped out window. Instead, you need to add the listener to `ev.view`, see for instance [1].
> 
> This is relatively easy for self-written code, it is more problematic in thirdparty libraries which may need patching.
> 
> [1] https://github.com/qgis/qwc2/blob/master/components/SideBar.jsx#

Thanks for the insight, I will look into it.
Maybe a less elegant way to solve this problem would be to destroy the viewer when the window pops out and reinitialize it in the new window.

I will try this weekend if I can.

tpo:
> > What does this plugin do? Are there screenshots of it or a screencast or documentation of it somewhere?
>
> The aim of this plugin is to implement support that allows users to consult images from a Panoramax instance in their QWC2 themes.
> 
> Panoramax is an open-source application that aims to centralize equirectangular (or other) images from all over the world. [...]

Many thanks @cazitouni for the explanation!

cazitouni:
Hello everyone,

Last commit adds the ability to load sequences as WMS, allowing users to view sequences even in a different projection.
However, Panoramax does not provide WMS capabilities, so users will need to host their own mirrors.

Since it's confirmed that OpenLayers does not support reprojecting MVT, I couldn't find a better solution for this issue.

I've set up a WMS mirror of api.panoramax.xyz [here](https://qgis.metrotopic.net/project/1e81ec32c524967f0e1f2f3497a12c24/?SERVICE=WMS&REQUEST=GetCapabilities).

Regarding the pop-out window issue, I still don't have a solution. After further investigation, it seems that the event listeners from Photo Sphere Viewer, the library used by the Panoramax viewer, remain attached to the original window. As a result, items in the popped-out window are no longer clickable, and the image moves when the cursor is moved over the map in the original window.

I'll try to look into this more when I have time, but my expectations for a fix are quite low.

Let me know if you have any ideas or suggestions!

manisandro:
Thanks, at a first glance the code looks good. Please add the proptype docstrings and also a comment block above the class, which will appear in the plugin documentation. Then I'd say it is good to merge, I'll have a look at the popout issue subsequently.

manisandro:
Thanks!

cazitouni:
Hello @manisandro, PR wasnt ready, I didn't have time to figure out how to get the assets from icons instead of `ConfigUtils.getAssetsPath()}/img` for the cursor and the indicator on the map.

Sorry for the inconvenience I'll look at it tonight.

PR #351, opened 2025-02-04
Calculate coordinates and bbox from feature

jblanchg:
Hi!

With the SearchBox and Search providers reworks our custom search provider stop working, I added this fix to calculate the marker coordinates and the bbox from the feature if the 'item' does not have it.


manisandro:
Is your search provider/backend able to return the search center / bbox directly in `onSearch`, or only when fetching the geometries via `getResultGeometry`?

jblanchg:
I can get the geometry for the item only via `getResultGeometry`

manisandro:
What about adding `center` and `bbox` to the `getResultGeometry` response, and using those instead if `item.{x,y,bbox}` if defined? I.e.

    CoordinatesUtils.reproject(response.center ?? [item.x, item.y], item.crs ?? this.props.map.projection, this.props.map.projection)

and

    let bbox = response.bbox ?? item.bbox ?? [item.x, item.y, item.x, item.y];

?

jblanchg:
Hi! I added bbox and center to the `getResultGeometry` if defined

manisandro:
Thanks

PR #353, opened 2025-02-10
Add global style selector button

jblanchg:
Hi! I'm not sure if this might be interesting to you, but since you can change the layer style individually from the LayerTree, I added an extra button to the title bar that allows the user to select a style and apply it to all layers (if the layer has that style available).

image

image

What do you thinkg?

Thanks!

### DIFF
--- a/plugins/LayerTree.jsx
+++ b/plugins/LayerTree.jsx
@@ -796,13 +796,13 @@ class LayerTree extends React.Component {
     };
 
     manageStyleSelector = () => {
-        let styleList = this.getLayerStyles(this.props.layers);
+        const styleList = this.getLayerStyles(this.props.layers);
         if (styleList.length > 0) {
-            this.setState({ styleList: styleList, activestyleselector: !this.state.activestyleselector });
+            this.setState((state) => ({styleList: styleList, activestyleselector: !state.activestyleselector}));
         }
     };
     getLayerStyles = (layerList) => {
-        let styleList = new Set();
+        const styleList = new Set();
         const collectStyles = (layers) => {
             for (const layer of layers) {
                 if (layer.sublayers && layer.sublayers.length > 0) {
@@ -810,7 +810,7 @@ class LayerTree extends React.Component {
                 }
                 if (layer.styles && Object.keys(layer.styles).length > 0) {
                     for (const styleKey in layer.styles) {
-                        if (layer.styles.hasOwnProperty(styleKey)) {
+                        if (Object.prototype.hasOwnProperty.call(layer.styles, styleKey)) {
                             styleList.add(layer.styles[styleKey]);
                         }
                     }
@@ -845,8 +845,8 @@ class LayerTree extends React.Component {
                 }
                 if (layer.styles && Object.keys(layer.styles).length > 0) {
                     for (const styleKey in layer.styles) {
-                        if (layer.styles.hasOwnProperty(styleKey) && styleKey === style) {
-                            const wmsLayer = this.props.layers.find(layer => layer.type === "wms");
+                        if (Object.prototype.hasOwnProperty.call(layer.styles, styleKey) && styleKey === style) {
+                            const wmsLayer = this.props.layers.find(l => l.type === "wms");
                             this.props.changeLayerProperty(wmsLayer.uuid, "style", style, adjustedPath);
                         }
                     }
--- a/plugins/LayerTree.jsx
+++ b/plugins/LayerTree.jsx
@@ -247,11 +247,16 @@ class LayerTree extends React.Component {
         }
         const optMenuClasses = classnames({
             "layertree-item-menubutton": true,
-            "layertree-item-menubutton-active": this.state.activemenu === group.uuid
+            "layertree-item-menubutton-active": this.state.activemenu === true
+        });
+        const styleMenuClasses = classnames({
+            "layertree-item-menubutton": true,
+            "layertree-item-menubutton-active": this.state.activestyleselector === group.uuid
         });
         const allowRemove = ConfigUtils.getConfigProp("allowRemovingThemeLayers", this.props.theme) === true || layer.role !== LayerRole.THEME;
         const allowReordering = ConfigUtils.getConfigProp("allowReorderingLayers", this.props.theme) === true && !this.state.filtervisiblelayers;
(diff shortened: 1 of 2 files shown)


manisandro:
Interesting! Would it also be an idea to place the selector on the toplevel WMS group layer (assumes you have "show root layer" enabled)? Rationale would be that, when adding additional themes, it might be desireable to switch styles for the themes separately - it is I guess pretty unlikely that separate themes will be sharing style names in general.

jblanchg:
Hi!

I applied the changes to add the button to the toplevel WMS group layer and to have the same style as the button to change the style of a layer individually.

image

image

plugins/LayerTree.jsx
```diff
@@ -787,6 +797,70 @@ class LayerTree extends React.Component {
...
+        }
+    };
+    getLayerStyles = (layerList) => {
+        const styleList = new Set();
+        const collectStyles = (layers) => {
+            for (const layer of layers) {
+                if (layer.sublayers && layer.sublayers.length > 0) {
+                    collectStyles(layer.sublayers);
+                }
+                if (layer.styles && Object.keys(layer.styles).length > 0) {
```
manisandro, line 817: This whole block could be shortened to

    const styleList = {};
    Object.assign(styleList, layer.styles);

plugins/LayerTree.jsx
```diff
@@ -246,9 +249,14 @@ class LayerTree extends React.Component {
             "layertree-item-menubutton": true,
             "layertree-item-menubutton-active": this.state.activemenu === group.uuid
         });
+        const styleMenuClasses = classnames({
+            "layertree-item-menubutton": true,
+            "layertree-item-menubutton-active": this.state.activestyleselector === group.uuid
```
manisandro, line 254: I believe you can re-use `activestylemenu` instead of adding a new `activestyleselector` state variable.

plugins/LayerTree.jsx
```diff
@@ -257,10 +265,12 @@ class LayerTree extends React.Component {
                     <span className="layertree-item-title" onClick={() => this.itemVisibilityToggled(layer, path, visibility)} title={group.title}>{group.title}</span>
                     {LayerUtils.hasQueryableSublayers(group) && this.props.allowSelectIdentifyableLayers ? (<Icon className={"layertree-item-identifyable " + identifyableClassName}  icon="info-sign" onClick={() => this.itemOmitQueryableToggled(layer, path, omitqueryable)} />) : null}
                     <span className="layertree-item-spacer" />
+                    {isTopLevelWMSGroup ? (<Icon className={styleMenuClasses} icon="paint" onClick={() => this.manageStyleSelector(group.uuid)}/>) : null}
```
manisandro, line 268: Ideally you'd show the menu only if styles are available. I'd collect the layer styles at this stage and only show the menu if styles are available. You can then directly pass the styles to the style selector rendering function, without storing them in the state.

plugins/LayerTree.jsx
```diff
@@ -787,6 +797,70 @@ class LayerTree extends React.Component {
...
+                            styleList.add(layer.styles[styleKey]);
+                        }
+                    }
+                }
+            }
+        };
+        collectStyles(layerList);
+        return Array.from(styleList);
+    };
+    renderStyleSelector = () => {
```
manisandro, line 829: You could potentially re-use `renderStyleMenu` by modifying the `renderStyleMenu` arguments to also take the list of styles to render as well as the callback.

plugins/LayerTree.jsx
```diff
@@ -787,6 +797,70 @@ class LayerTree extends React.Component {
...
+                    for (const styleKey in layer.styles) {
+                        if (Object.prototype.hasOwnProperty.call(layer.styles, styleKey) && styleKey === style) {
+                            const wmsLayer = this.props.layers.find(l => l.type === "wms");
+                            this.props.changeLayerProperty(wmsLayer.uuid, "style", style, adjustedPath);
+                        }
+                    }
+                }
+            }
+        };
+        setStyle(this.props.layers);
```
manisandro, line 862: You only want to apply the style to the one layer, not all layers.

jblanchg:
Thanks for the feedback!

I tried to reuse both `activestylemenu `and `renderStyleMenu `

plugins/LayerTree.jsx
```diff
@@ -257,10 +264,12 @@ class LayerTree extends React.Component {
...
                     {LayerUtils.hasQueryableSublayers(group) && this.props.allowSelectIdentifyableLayers ? (<Icon className={"layertree-item-identifyable " + identifyableClassName}  icon="info-sign" onClick={() => this.itemOmitQueryableToggled(layer, path, omitqueryable)} />) : null}
                     <span className="layertree-item-spacer" />
+                    {isTopLevelWMSGroup && Object.keys(styles).length > 0 ? (<Icon className={styleMenuClasses} icon="paint" onClick={() => this.layerStyleMenuToggled(group.uuid)}/>) : null}
                     <Icon className={optMenuClasses} icon="cog" onClick={() => this.layerMenuToggled(group.uuid)}/>
                     {allowRemove ? (<Icon className="layertree-item-remove" icon="trash" onClick={() => this.props.removeLayer(layer.id, path)}/>) : null}
                 </div>
                 {this.state.activemenu === group.uuid ? this.renderOptionsMenu(layer, group, path, allowRemove) : null}
+                {isTopLevelWMSGroup && this.state.activestylemenu === group.uuid ? this.renderStyleMenu(styles, this.state.selectedStyle, (style) => this.applyLayerStyle(style)) : null}
```
manisandro, line 272: The `selectedStyle` state variable will not work correctly if there are multiple toplevel groups with different styles. Instead, look at layer.params.STYLES and check if it contains one unique or different values to decide which style to mark as selected.

jblanchg: Thanks! I think it should work ok now

plugins/LayerTree.jsx
manisandro: This still needs to be addressed.

jblanchg: You're right, sorry! On the latest commit I pass the toplevel layer to `applyLayerStyle` directly.

manisandro:
Thanks!

PR #355, opened 2025-02-15
Japanese translation corrected.

softark:
Corrected Japanse translation for "search.placeholder".

Its counterpart in 'en-US.json' is currently "Search place or add map...", but I don't think it's suitable for the use case in 'searchBox' in the 'TopBar'. So I've given a Japanese translation for something like "Search themes and layers to select or add...".

Or, "search.placeholder" has other use cases ? Then tell me so, and I will withdraw the PR.

### DIFF
--- a/translations/ja-JP.json
+++ b/translations/ja-JP.json
@@ -506,7 +506,7 @@
       "nodescription": "説明がありません",
       "none": "無し",
       "noresults": "検索結果無し",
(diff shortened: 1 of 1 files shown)


softark:
OK, I've got it. I'm sorry, I was wrong. 
The search box really searches "places" when you have correctly configured the search provider of the theme.
So I withdraw this PR.  Thanks.

PR #356, opened 2025-02-17
Enhance WMTS imports by handling styles and tile matrices per layer

benoitblanc:
Hi,

Currently, I do not manage to use some WMTS services correctly :
* Grand Lyon (https://imagerie.data.grandlyon.com/geoserver/grandlyon/gwc/service/wmts?REQUEST=GetCapabilities)
* Grand Est (https://www.datagrandest.fr/geoserver/gwc/service/wmts?REQUEST=GetCapabilities)
* IGN (https://data.geopf.fr/wmts?SERVICE=WMTS&VERSION=1.0.0&REQUEST=GetCapabilities)

Requests are not the right ones because `TileMatrix` parameter is wrong due to `TileMatrixSet.TileMatrix.Identifier`

For instance, when adding a layer from Grand Lyon using the demo app : https://imagerie.data.grandlyon.com/geoserver/grandlyon/gwc/service/wmts?layer=TS2011_nuit&style=raster&tilematrixset=EPSG%3A4326&Service=WMTS&Request=GetTile&Version=1.0.0&Format=image%2Fjpeg&TileMatrix=21&TileCol=19&TileRow=4 raises an error `Unknown TILEMATRIX 21`

When requesting GetCapabilities, TileMatrixSet.TileMatrix.Identifier is not a simple incremental number but `EPSG:3946:0` so we have to handle this prefix to build the request.

Also, there could be multiple styles, multiple tile matrices and formats by layer. So the layer currently displayed in the app right now could not be working.

There was also an issue regarding resolutions that have to be computed regarding max resolution (see [OpenLayers example on WMTS](https://openlayers.org/en/latest/examples/wmts.html)).

I tried to manage multiple styles and tile matrices with `sublayers` to be expanded. For now, I search for `image/jpeg` format or fallback to the first format returned in capabilities. This could be improved later if we want to handle multiple formats.

image

image

The default preset (Swisstopo https://wmts10.geo.admin.ch/EPSG/2056/1.0.0/WMTSCapabilities.xml) is still working with this adding.

Maybe there are still remaining issues on specific cases (i.e EPSG:4326 that is not working, and maybe other layers) but I think the situation is better than it was regarding the 3 french examples that I used. Specific cases could be treated later.

Please tell me what you think about it.

Thanks

Benoît

### DIFF
--- a/components/map/layers/WMTSLayer.js
+++ b/components/map/layers/WMTSLayer.js
@@ -30,7 +30,7 @@ export function createWMTSSource(options) {
     const matrixIds = new Array(options.resolutions.length);
     // generate matrixIds arrays for this WMTS
     for (let z = 0; z < options.resolutions.length; ++z) {
-        matrixIds[z] = options.tileMatrixPrefix + z;
+        matrixIds[z] = options.tileMatrixPrefix !== "" ? options.tileMatrixPrefix + ":" + z : z;
     }
     const extent = options.bbox ? CoordinatesUtils.reprojectBbox(options.bbox.bounds, options.bbox.crs, options.projection) : null;
 
--- a/utils/ServiceLayerUtils.js
+++ b/utils/ServiceLayerUtils.js
@@ -10,6 +10,8 @@ import axios from 'axios';
 import deepmerge from 'deepmerge';
 import {XMLParser} from 'fast-xml-parser';
 import isEmpty from 'lodash.isempty';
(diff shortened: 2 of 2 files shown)


manisandro:
Thanks for the PR. I think best is to merge it and then collect feedback. Regarding `format`, perhaps `png` is preferrable due to potential transparency?

manisandro:
Thanks!

PR #360, opened 2025-02-26
Fix validateGeometry for GeometryCollection

cazitouni:
Hello @manisandro,

I made this pull request related to https://github.com/qgis/qwc2-demo-app/issues/606.
I've tested it with MultiSurface and MultiPolygon geometries, and it seems to work fine with no apparent side effects.

Let me know if there is something else.

Regards, Clément.


manisandro:
Thanks!

PR #362, opened 2025-03-05
fixed logged in user name z-index

softark:
As logged-in user name widget had no z-index setting, it was hidden behind the map image.
Now it has the same z-index as map buttons and hovers over the map image.

Corresponding issue in `qwc2-demo-app`
- https://github.com/qgis/qwc2-demo-app/issues/608


manisandro:
Thanks!

PR #393, opened 2025-03-07
ja-JP translation for "redlining" revised

softark:
Revided the translation of "Redlining".
Originally I used the literal translation 「赤 (red) 線 (line) 引き (drawing)」, but 「赤線」in Japanese has a meaning of "prostitution street" in general, and a technical meaning of "public road lines" in geospacial profession.
So I changed it from 「赤線引き」 to 「スケッチ」, i.e., "Sketch(ing)".


manisandro:
Thanks!

PR #394, opened 2025-03-07
Panoramax assets and configuration

cazitouni:
Regarding the end of https://github.com/qgis/qwc2/pull/348 and https://github.com/qgis/qwc2-demo-app/pull/61. 

I didn't realize you have merged the core and the demo app, so i've added the Panoramax related assets directly into the core, same goes for the configuration in Appconfig.js. 

let me know if it works for you. 

Regards, Clément.


manisandro:
Thanks

Issue #399, opened 2025-03-28
No BBOX sent when using Identify Region


cazitouni:
The Identify Region plugin does not send BBOX parameters in its GetFeatureInfo request. 

For example : 

`https://viewer.wisiglw.cus.fr/api/v1/featureinfo/canard?service=WMS&version=1.3.0&request=GetFeatureInfo&id=4e9cb498-9c29-4044-b5b7-95eb258e32c0&layers=vo_eqr_totem,chien&query_layers=vo_eqr_totem,chien&srs=EPSG:3948&crs=EPSG:3948&info_format=text%2Fxml&with_geometry=true&with_maptip=false&height=101&width=101&feature_count=100&FILTER_GEOM=POLYGON+((2050420.0854+7276176.9214,+2050421.2099+7276170.2407,+2050432.1901+7276169.4469,+2050430.6687+7276178.3766,+2050420.0854+7276176.9214))&FI_LINE_TOLERANCE=8&FI_POINT_TOLERANCE=16&FI_POLYGON_TOLERANCE=4`

The problem is that without a BBOX, it selects features even if their layer is not rendered at the current scale.
The Identify Point plugin does not have this issue, as it sends the map's BBOX with the request.

`https://viewer.wisiglw.cus.fr/api/v1/featureinfo/canard?service=WMS&version=1.3.0&request=GetFeatureInfo&id=4e9cb498-9c29-4044-b5b7-95eb258e32c0&layers=vo_eqr_totem,chien&query_layers=vo_eqr_totem,chien&srs=EPSG:3948&crs=EPSG:3948&info_format=text%2Fxml&with_geometry=true&with_maptip=false&height=101&width=101&feature_count=20&x=51&y=51&i=51&j=51&bbox=2050420.449207053,7276170.472178425,2050427.1299362197,7276177.152907592&FI_LINE_TOLERANCE=8&FI_POINT_TOLERANCE=16&FI_POLYGON_TOLERANCE=4&region_feature_count=100`

Using QGIS Server 3.42, simply adding the BBOX from the second request to the first one seems to work.
I'm not sure if this is intentional. If not, I might consider submitting a PR with a fix.

Let me know. 
Regards, Clément.

manisandro:
I guess this is just an oversight. PR welcome!

cazitouni:
Fixed by #401

PR #401, opened 2025-03-31
Fix missing BBOX parameter in Identify region

cazitouni:
Hello @manisandro,

This update addresses the missing BBOX parameter in the Identify Region plugin (#399) and other plugins such as snapping, reporting, TimeManager, and PickFeature.

I believe it's better to add the parameter globally since, ultimately, you want to select only what is visible on the map—but I might be wrong.

Please let me know if there are any issues.

Best regards,
Clément


manisandro:
Shouldn't harm to set it there, thanks!

PR #403, opened 2025-04-01
Set version to 2025.03.31

benoitblanc:
Hello @manisandro, is it possible to push a tag `v2025.03.31` so the qwc-map-viewer image can be built successfully please ?

Thanks


manisandro:
Sure, I've pushed the tag

PR #412, opened 2025-04-25
Allow to disable themes in configuration

benoitblanc:
Hi,

This PR allows user to disable temporarily some themes in configuration to be able to access to the application despite some themes not well configured (bad layers, external layers not available, ...).

I will create an other PR in qwc-config-generator and qwc-admin-gui services to handle this feature.

This is funded by [Eurométropole de Strasbourg](https://www.strasbourg.eu/).

Thanks for the review


manisandro:
Thanks!

PR #413, opened 2025-04-29
Add option to print highlights into map

BenediktSeidlSWM:
This pull request adds a simple option `printMapHighlights` to the Print plugin which can be used to disable the printing of highlighted features or additional redlining.


manisandro:
Thanks!

PR #416, opened 2025-05-02
Filter expression in comboBox of attribute form

rsrg-zwiama:
The filter expressions are stored in the editConfig and not in the fields in the mapViewerConfig. So the filter expression configured in qgis relational fields are used in qwc, when loading from editConfig (like in the AttributeTable, where it is also loaded from the editConfig).


manisandro:
Thanks - you could pass down the `this.props.theme.editConfig` from the parent `AttributeForm` component to avoid having to observe `theme` also in the `QtDesignerForm` component.

manisandro:
Thanks!

rsrg-zwiama:
@manisandro Are your planing to merge this into the lts 2025?

manisandro:
I've cherry-picked it to 2025-lts

manisandro:
@rsrg-zwiama I had to revert this as it caused the application to crash with simple kvrels (not in a relation)

PR #419, opened 2025-05-08
Fix MVT styles

cazitouni:
Hello @manisandro, 

I allow myself to submit this PR in order to fix a bug that happens if the MVT style does not contains gylph or sprites. 

For example : 

https://api.panoramax.xyz/api/map/style.json

Let me know if this works for you. 
Regards, Clément.


manisandro:
Thanks!

PR #421, opened 2025-05-12
Update panoramax web viewer to v4

cazitouni:
Hello @manisandro,

I am submitting this pull request to update the Panoramax Plugin to the new version 4.

Panoramax has updated their web viewer to a new major version, which introduced many changes to the code. As a result, I had to rewrite the component somewhat.

I have tested it locally, and everything seems fine. However, please feel free to give your opinion and review the code before merging.

Kind regards,
Clément

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
         "@loaders.gl/core": "^4.3.3",
         "@loaders.gl/shapefile": "^4.3.3",
         "@loaders.gl/zip": "^4.3.3",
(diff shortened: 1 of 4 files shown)


manisandro:
Thanks, I suppose yarn.lock should also be updated?

cazitouni:
Ok, I was thinking the package file was enough, here it is with the lock file updated with an install.

manisandro:
Thanks!

PR #422, opened 2025-05-14
filterExpression comboField with attrname currentLayerId

rsrg-zwiama:
the currentEditConfig is defined different for tables and reltables. Current solution only works for reltables

### DIFF
--- a/components/QtDesignerForm.jsx
+++ b/components/QtDesignerForm.jsx
@@ -390,20 +390,26 @@ class QtDesignerForm extends React.Component {
                 // kvrel__attrname__datatable__keyfield__valuefield
                 // kvrel__reltablename__attrname__datatable__keyfield__valuefield
                 const count = parts.length;
-                const attrname = parts.slice(1, count - 3).join("__");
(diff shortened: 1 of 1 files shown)


manisandro:
Could you give some additional details regarding the editConfig difference you mention?

rsrg-zwiama:
@manisandro sure, I try to explain.
The goal is to read the filter expression in the editConfig. The editConfig contains the details for every layer of the project. So fist, I select the correct dataset. If I have a simple edit layer with a combobox selection, the dataset name in the editConfig corresponds to the editLayerId. If i have a relational table with a combobox, the editLayerId is the layer of the 'parent'. I can create/read the name of the layer/tablename from the widget (const parts = widget.name.split("__");) . So i added the extra variable currentLayerId, which I use to select the correct dataset from editConfig into currentEditConfig. The currentLayerId depends on the table type (normal table/layer or child/ralation table).

manisandro:
I believe https://github.com/qgis/qwc2/pull/416 needs to be revisited before this one.

PR #423, opened 2025-05-14
Fix undefined formatConfiguration

BenediktSeidlSWM:
The bug also occurs on the qwc2.sourcepole.ch demo instance, where map export is currently broken.


manisandro:
Thanks!

PR #424, opened 2025-05-14
Add option whether to show a highlight marker for the search results

BenediktSeidlSWM:
The setting defaults to true as this is the established behavior.


manisandro:
So if this is set to false, a regular point geometry is shown, right?

BenediktSeidlSWM:
Yes exactly, that is the desired layout for our use case.

manisandro:
Ok, thanks!

PR #429, opened 2025-05-26
Update Print.jsx

hirtarmin:
Prevents certain print layouts from being displayed in the layout list of the print dialog.

Several print layouts are defined for a QWC theme. Some of them shouldn't be displayed in the layout list of the print dialog, but are used to print an info portfolio site (special reports created with a custom service).
Therefore, I can't exclude these print layouts in QGIS.

So I defined a prefix for print layout names. All layouts starting with these prefix are not displayed in the print dialog.
The prefix is ​​specified in the config.json file. I also specified a sort order and a default layout:

Added plugin config variables:
1. hidePrintlayoutPrefix: All layouts with this prefix will not be displayed in the print layoutlist
2. sortPrintlayouts: Alphabetical order for print layouts: asc or desc
3. defaultPrintlayout: The default layout to display first

config.json (example):
{
  "name": "Print",
  "cfg":  #{
    …
    "hidePrintlayoutPrefix": "_hide",
    "sortPrintlayouts": "desc",
    "defaultPrintlayout": "A4 - Hochformat"
  }
}

### DIFF
--- a/plugins/Print.jsx
+++ b/plugins/Print.jsx
@@ -86,9 +86,7 @@ class Print extends React.Component {
         /** All layouts with this prefix will not be displayed in the print layoutlist */
         hidePrintlayoutPrefix: PropTypes.string,
         /** Alphabetical order for print layouts: asc or desc */
-        sortPrintlayouts: PropTypes.string,
-        /** The default layout to display first */
-        defaultPrintlayout: PropTypes.string
+        sortPrintlayouts: PropTypes.string
     };
     static defaultProps = {
         defaultDpi: 300,
@@ -103,8 +101,7 @@ class Print extends React.Component {
         printMapHighlights: true,
         scaleFactor: 1.9, // Experimentally determined...
         side: 'right',
-        sortPrintlayouts: 'desc',
-        defaultPrintlayout: ''
+        sortPrintlayouts: 'desc'
     };
     state = {
         center: null,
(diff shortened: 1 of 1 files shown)


plugins/Print.jsx
```diff
@@ -83,6 +83,12 @@ class Print extends React.Component {
         /** The side of the application on which to display the sidebar. */
         side: PropTypes.string,
         theme: PropTypes.object
+        /** All layouts with this prefix will not be displayed in the print layoutlist */
+        hidePrintlayoutPrefix: PropTypes.string,
+        /** Alphabetical order for print layouts: asc or desc */
+        sortPrintlayouts: PropTypes.string,
+        /** The default layout to display first */
+        defaultPrintlayout: PropTypes.string
```
manisandro, line 91: There already exists the option `defaultPrintLayout` in the theme configuration item for this [1]

[1] https://qwc-services.github.io/master/configuration/ThemesConfiguration/#manual-theme-configuration

plugins/Print.jsx
```diff
@@ -237,7 +255,7 @@ class Print extends React.Component {
                             <td>{LocaleUtils.tr("print.layout")}</td>
                             <td>
                                 <select onChange={this.changeLayout} value={this.state.layout.name}>
-                                    {this.state.layouts.map(item => {
+                                    {this.state.layouts.filter(item => item.name.startsWith(this.props.hidePrintlayoutPrefix)).map(item => (
```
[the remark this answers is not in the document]
hirtarmin: Sorry, there was still a line of original code... Thanks for correcting it.

danceb:
As the print layouts are configured per theme and there some other theme related config parameters like the mentioned `defaultPrintLayout` and others like `printLabelConfig`, I would suggest to set the config parameters within the `themesConfig.json` and not the `config.json`. 
If they should be set to all of the themes (i.e. in combination with using the QGIS Server Plugin "print_templates" [1]) this should be done with default Parameters like `defaultHidePrintlayoutPrefix` and `defaultSortPrintLayouts`.

[1] https://qwc-services.github.io/master/topics/Printing/#layout-templates

plugins/Print.jsx
hirtarmin: I made an additional commit to remove the intended `defaultPrintlayout` option.

Issue #409, opened 2025-04-22
Plugin suggestion -  Tour guide


cazitouni:
Hello @manisandro,

I’d like to open this ticket to share an idea regarding a potential Tour Guide plugin.

While the application is simple and intuitive, we've noticed that some of our less tech-savvy users still struggle to find and use basic features.

To help with this, I was thinking of implementing a Tour Guide plugin using a library such as:

- https://www.onborda.dev/
- https://driverjs.com/

The idea would be to add a button that launches a step-by-step tour covering the essential features of QWC. I believe this could greatly improve usability, especially for new users.

I'll probably implement something like this for our intranet client regardless, but I’d love to hear your thoughts on whether this could also be useful for the standard client.

Kind regards,
Clément

PR #432, opened 2025-05-29
DriverJS TourGuide

cazitouni:
Hello @manisandro,

I'm opening this PR in relation to issue #409.

I've implemented a component using the [DriverJS library](https://driverjs.com/) to provide a guided tour feature. The component uses a `.json` file as its configuration, which can be loaded either from the static files or from a URL. I’ve included a basic example JSON file to demonstrate the component's capabilities.

The component also includes methods to open menus and load tasks, allowing them to be highlighted during the tour. This works in most cases, though there are exceptions—such as the TimeManager plugin, which doesn't use the `setTask(null)` method for unloading.

I'd appreciate your feedback on the implementation. If everything looks good, I'll proceed to add documentation, especially outlining the JSON configuration options.

You can find the tour guide in the menu above the Help section in this [live demo](https://qwc.metrotopic.net).

Kind regards,  
Clément


manisandro:
Thanks, very interesting! One slight improvement: if the DriverJS library is not trivially small, could you delay-load it? See i.e. [1].

[1] https://github.com/qgis/qwc2/blob/master/plugins/GeometryDigitizer.jsx#L340

cazitouni:
Hello @manisandro, thanks for the merge. 

In order to document how the config .json works for the tour guide. Should I just make a new page in https://github.com/qwc-services/qwc-services.github.io ? 

Thanks !

PR #439, opened 2025-06-08
Added defaultDisplayCrs option for theme config

longerCZE:
As discussed in issues I have added option defaultDisplayCrs for both theme and themes with the same name. I have also modified themeconfig scripts.


manisandro:
Thanks! Could you also update the documentation at [1]?

[1] https://github.com/qwc-services/qwc-services.github.io/blob/master/src/configuration/ThemesConfiguration.md

PR #440, opened 2025-06-12
Reset extra info for new mapinfo coordinates

BenediktSeidlSWM:
When selecting a new point to show the mapinfo tooltip, the previous information is still displayed. This behavior can be quite confusing, especially when the correct data takes some time to load.


manisandro:
Thanks!

PR #443, opened 2025-06-13
Adds display options for coordinates

longerCZE:
Also PR with documentation update on the way...


longerCZE:
I have messed repo when playing with eslint, will create new PR.

PR #446, opened 2025-06-16
set layer visibility recursively

stephaschu:
When a layer is inside a group, the group is currently not activated.


manisandro:
Thanks!

PR #450, opened 2025-06-25
Allow choosing inteval units on routing isochrones

NullSeile:
Added support for multiple customizable time and distance units for the intervals on the routing plugin.

We have added by default minuts and seccons for time, and kilometers and meters for distance.


manisandro:
Thanks!

PR #469, opened 2025-07-27
Added support for displaying coordinates options like in QGIS

longerCZE:
After some vacation break i have finally managed to finish this, sorry for delay..


components/CoordinateDisplayer.jsx
```diff
@@ -31,16 +31,18 @@ export default class CoordinateDisplayer extends React.Component {
...
+            );
         }
         return (
-            <input className={this.props.className} readOnly="readOnly" type="text" value={value}/>
+            <input className={this.props.className} readOnly type="text" value={value} />
         );
     }
     getMapMousePos = (ev) => {
-        this.setState({mousePos: ev.coordinate});
+        this.setState({ mousePos: ev.coordinate });
```
manisandro, line 46: Please avoid whitespace changes

manisandro:
See [1] for changes you need to apply. Or just use the qwc2 repository directly, the qwc2-demo-app is actually nowadays only for advanced uses.

[1] https://qwc-services.github.io/master/release_notes/QWC2UpgradeNotes/

PR #472, opened 2025-08-02
Add options to set coordinates display format as in QGIS

longerCZE:
New version which we have discussed in PR https://github.com/qgis/qwc2/pull/469
Documentation here: https://github.com/qwc-services/qwc-services.github.io/pull/30


manisandro:
Thanks!

PR #476, opened 2025-08-06
render multiple height profiles

rsrg-zwiama:
this PR refers to: https://github.com/qwc-services/qwc-elevation-service/issues/17 
I changed the HeightProfile plugin, so multiple dsm data will render parallel in the height profile window. The datasets can be activated my check boxes. Datasets with only 0 elevation values (when line outside of dsm) are ignored and also the minimum of the y-scale is changed, so the 0 elevation are not used to calculate the min value. The tooltip shows both dsm values, when hovering over the line in the map and one value when hovering in the profile window.


plugins/HeightProfile.jsx
```diff
@@ -300,19 +300,23 @@ class HeightProfile extends React.Component {
...
-                return {
-                    dataset: entry.dataset,
-                    x: x,
-                    y: elevations,
-                    maxY: Math.max(...elevations),
-                    totLength: totLength,
-                    nodes: nodes
-                };
-            });
+                if (!elevations.every(val => val === 0)) {
```
manisandro, line 303: Since you are computing `Math.min(...elevations.filter(elev => elev !== 0))` below, you might as well do
```
nonZeroElevations = elevations.filter(elev => elev !== 0)
if (nonZeroElevations.length > 0) {
...
minY: Math.min(...nonZeroElevations),
...
}
```

Saves off one iteration over the data.

plugins/HeightProfile.jsx
```diff
@@ -300,19 +300,23 @@ class HeightProfile extends React.Component {
...
+                    return {
+                        dataset: entry.dataset || `Dataset ${index + 1}`,
```
manisandro, line 305: The fallback should be localized.

plugins/HeightProfile.jsx
```diff
@@ -445,36 +457,47 @@ class HeightProfile extends React.Component {
...
-                        </option>
-                    ))}
-                </select>
+                <div className="height-profile-dataset-select">
+                    {this.state.data.map((dataset, idx) => {
+                        const isSelected = this.state.selectedDatasetIndices?.includes(idx);
+                        return (
+                            <div key={`${dataset.dataset || 'unnamed'}-${idx}`} >
+                                <span
+                                    className={`icon icon_clickable ${ isSelected ? 'icon-checked' : 'icon-unchecked' } `} onClick={() => {
```
manisandro, line 480: Better to use the `Icon` component instead of `span className="icon"`

plugins/HeightProfile.jsx
```diff
@@ -445,36 +457,47 @@ class HeightProfile extends React.Component {
...
+                                        const selected = new Set(this.state.selectedDatasetIndices || []);
+                                        if (isSelected) {
+                                            selected.delete(idx);
+                                        } else {
+                                            selected.add(idx);
+                                        }
+                                        this.setState({ selectedDatasetIndices: Array.from(selected) });
```
manisandro, line 487: This does not guarantee that at least one index is selected?

plugins/HeightProfile.jsx
```diff
@@ -529,7 +555,7 @@ class HeightProfile extends React.Component {
             this.clearMarkerAndTooltip();
             return;
         }
-        const data = this.state.data[this.state.selectedDatasetIndex];
+        const data = this.state.data[this.state.selectedDatasetIndices[0]];
```
manisandro, line 558: This assumes that at least one index is selected?

manisandro:
Thanks, I've added some review comments

plugins/HeightProfile.jsx
manisandro: Yes, or we could actually force via elevation service config schema that a dataset is named.

plugins/HeightProfile.jsx
```diff
@@ -445,36 +457,47 @@ class HeightProfile extends React.Component {
                         display: true,
                         text: heightStr + " [m " + aslStr + "]"
                     },
-                    max: Math.ceil(selectedDataset.maxY)
+                    max: Math.ceil(Math.max(...selectedDatasetIndices.map(idx => this.state.data[idx]?.maxY ?? 0))),
+                    min: Math.floor(Math.min(...selectedDatasetIndices.map(idx => this.state.data[idx]?.minY ?? 0)))
```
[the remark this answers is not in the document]
rsrg-zwiama: yes, I remove them

plugins/HeightProfile.jsx
```diff
@@ -445,36 +457,47 @@ class HeightProfile extends React.Component {
...
+                                        const selected = new Set(this.state.selectedDatasetIndices || []);
+                                        if (isSelected) {
+                                            selected.delete(idx);
+                                        } else {
+                                            selected.add(idx);
+                                        }
+                                        this.setState({ selectedDatasetIndices: Array.from(selected) });
+                                    }}
+                                />
+                                <span>{dataset.dataset || "<unnamed>"}</span>
```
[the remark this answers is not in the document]
rsrg-zwiama: yes, that's from the old version. I will remove unnamed

plugins/HeightProfile.jsx
rsrg-zwiama: Yes it is possible to deselect all datasets. In this case, the chart will show no data (profile line). I added fallback values, so it should work without selected dataset.

plugins/HeightProfile.jsx
rsrg-zwiama: It is possible to unselec all. so I added fallback here

Issue #486, opened 2025-09-01
support import layer from KMZ


4xoc:
Currently only a few formats are supported when creating a new layer. While KML is already supported, KMZ isn't. Can support be added for KMZ too? It looks like that would just require to unzip the file and extract the KML inside.

Additionally when trying to add an unsupported file one gets no error message whatsoever. The layer just isn't created but for the user there is no indication what and why went wrong.

manisandro:
Displaying a warning is a fair point, implemented in [1]. Regarding KMZ, adding something like this [2] should be pretty easy. Do you want to give it a shot? Look at [3][4][5][6].

- [1] https://github.com/qgis/qwc2/commit/83af10b45363306a9e99a7e7e09e302c06e609bb
- [2] https://openlayers.org/en/latest/examples/drag-and-drop-custom-kmz.html
- [3] https://github.com/qgis/qwc2/blob/83af10b45363306a9e99a7e7e09e302c06e609bb/components/ImportLayer.jsx#L57
- [4] https://github.com/qgis/qwc2/blob/83af10b45363306a9e99a7e7e09e302c06e609bb/components/ImportLayer.jsx#L248
- [5] https://github.com/qgis/qwc2/blob/83af10b45363306a9e99a7e7e09e302c06e609bb/components/ImportLayer.jsx#L270
- [6] https://github.com/qgis/qwc2/blob/83af10b45363306a9e99a7e7e09e302c06e609bb/utils/VectorLayerUtils.js#L437

PR #489, opened 2025-09-08
feat(#486): support import of kmz layer

4xoc:
This PR adds support for importing KMZ layers as discussed in https://github.com/qgis/qwc2/issues/486.


manisandro:
Thanks!

PR #496, opened 2025-09-25
Set version to 2025.09.24

benoitblanc:
This is to test new NPM package with resources.

Thanks


manisandro:
Is now tagged, sorry for delay!

PR #502, opened 2025-10-01
feat: add shapefile export

dominik-warch:
Adds a new Esri Shapefile export option to the IdentifyViewer component.

Changes:
* new CoordinateUtils.getWktFromCrs() method to convert EPSG codes to WKT
* new shapefile exporter in IdentifyViewer.jsx with single/multi-layer support (based on existing CSV and GeoJSON export structure)
* add @mapbox/shp-write dependency for shapefile generation
* add @esri/proj-codes for WKT/CRS conversion

I'm a bit concerned about the @esri/proj-codes dependency, as the package is  ~16 MB in size.

An alternative would be to maintain a simple lookup table with frequently used ESPGs (4326, 3857, etc.) and the option to add your own via JSON.

### DIFF
--- a/components/IdentifyViewer.jsx
+++ b/components/IdentifyViewer.jsx
@@ -16,6 +16,7 @@ import JSZip from 'jszip';
 import isEmpty from 'lodash.isempty';
 import omit from 'lodash.omit';
 import PropTypes from 'prop-types';
+import shpwrite from '@mapbox/shp-write';
 
 import {setActiveLayerInfo} from '../actions/layerinfo';
 import {LayerRole, addLayerFeatures, removeLayer, changeLayerProperty} from '../actions/layers';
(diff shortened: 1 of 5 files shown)


manisandro:
Yes latter seems the best option, perhaps adding a warning / console log statement if when attempting to export and no wkt is defined.

dominik-warch:
Removed @esri/proj-codes and added a WKT lookup using common WKTs in the CoordinatesUtils itself as fallback after searching in the config.json.

For the provided projections in the config.json I added the corresponding WKT. I named them esriWkt because Esri uses its own WKT codes and not the official ones from OGC.

components/IdentifyViewer.jsx
```diff
@@ -152,6 +152,74 @@ const BuiltinExporters = [
...
+                const layers = Object.entries(json);
+                const options = {
+                    outputType: 'arraybuffer',
+                    types: {
+                        point: 'points',
+                        polygon: 'polygons',
+                        polyline: 'lines'
+                    }
+                };
+                if (layers.length === 1) {
```
manisandro, line 171: To reduce code duplication, you could directly do

    const promises = layers.map(([layerName, features]) => { .... });

and then at the end, depending on the number of layers, decide whether to zip the result or return the single shp zip directly.

dominik-warch:
We could also define the excluded properties in a constant and reuse it in the geojson and shapefile export.

```js
const EXCLUDE_PROPS = ['featurereport', 'displayfield', 'layername', 'layertitle', 'layerinfo', 'attribnames', 'clickPos', 'displayname', 'bbox']

...

const feature = omit(entry, EXCLUDE_PROPS);
```

manisandro:
Thanks!

PR #504, opened 2025-10-02
Introduce setting to move print series

### LINKED ISSUE #448: Moving print extent in print series
@BenediktSeidlSWM I will adress this to you, as the PR https://github.com/qgis/qwc2/pull/340 was introduced by you.

Within the reworked print plugin the print extend could only be moved in single mode (print series disabled). Indeed it would be nice to move the print extend (and the other series sheets) aswell, if the print series mode is activated.

May you implement this functionality? I think, it could work the same way as in single mode. Click and drag within the active sheets of the grid should move the whole grid and click and drag outside of them should move the map view in the background.

What do you think about it?

BenediktSeidlSWM:
Fixes #448


manisandro:
Thanks!

PR #519, opened 2025-11-03
Attribute form size nestedforms

rsrg-zwiama:
when using nested forms for relational values, the form renders on top of the edit window and gets the height of the edit window. Since there is not much content in the base edit window, the window height is small and the form doesn't scale to the form height. With this changes, the overlaying form pushes the window height to the attribute form style height, when the form is shown and removed when form editing is finished.

### DIFF
--- a/components/AttributeForm.jsx
+++ b/components/AttributeForm.jsx
@@ -60,13 +60,11 @@ class AttributeForm extends React.Component {
         childEdit: null,
         relationTables: {},
         formValid: true,
-        captchaResponse: null,
-        dynamicHeight: false
+        captchaResponse: null
     };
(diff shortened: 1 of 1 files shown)


manisandro:
Thanks, could you post a before/after screenshot for better understanding?

rsrg-zwiama:
Hi @manisandro Now I missed your response for quite a while. I tested the behavior and can say, that the changes from the parent form are not lost, when opening a child form and return to the parent. But what I realized is, that it always opens the first tab of the parent form, when closing the child form. And I realized, that i can not create new child features with geometry. 

So You mean, that I do a css selector depending on this.state.childEdit? I will give it a try.

PR #520, opened 2025-11-03
feat: copy matching attribute values during clone geometry action

dominik-warch:
When using "Clone from geometry" in the Editing plugin, matching attribute values (identical id) are now copied from the source feature to the new feature as well.


dominik-warch:
Good idea! I tested an implementation using the MessageBar component:

image

While testing with different layers, I encountered the following situation: Currently, it copies all attributes for which there is a match, regardless of the visibility in the edit form. The number shown in the example (12) also refers to the total number of copied attributes. From a UX perspective, however, it is confusing when (as in my example) only six visible fields have actually copied values.

Off the top of my head, I can think of the following options:
* Limit copying to only visible elements
* Add to the dialog, e.g., “Found 12 matching attributes, 6 visible.” or something like this.

manisandro:
A somewhat elaborate solution could be to display matching attributes as a list with checkboxes in a dialog, then allowing the user to select "none", "only visible", "all", or a custom selection.

manisandro:
It should contain `hidden` in the constraints, i.e.

image


If not, this would need to be investigated in the config-generator code [1].

[1] https://github.com/qwc-services/qwc-config-generator/blob/master/src/config_generator/qgs_reader.py#L588

dominik-warch:
Ah, I see! Hidden references to the widget type. But when I build my form with the drag-and-drop editor, I also need to set the widget type for the available fields in the left column, not only for the ones actually used in the right column. The editConfig exposes all fields, regardless of whether I used them in the layout or not.

image

manisandro:
Yes that is indeed the case, because the field configuration in the editConfig is also used for the attribute table.

dominik-warch:
I added your suggested checkbox selection. By default hidden and calculated fields are deselected.

image

plugins/Editing.jsx
```diff
@@ -142,7 +143,50 @@ class Editing extends React.Component {
             this.setState({pickedFeatures: null});
         }
     }
+    renderCloneAttributeDialog = () => {
+        const quickButtons = [
+            {key: 'All', label: LocaleUtils.tr("editing.clone_all")},
+            {key: 'None', label: LocaleUtils.tr("editing.clone_none")},
+            {key: 'Visible', label: LocaleUtils.tr("editing.clone_visible")}
```
manisandro: You could store whether there are any hidden attributes when building `pendingClone` below, and display this button only whether there are indeed hidden attributes. Possibly you could render the hidden attributes italic in the list as a hint to the user?

dominik-warch: Currently I add "hidden" or "calculated" in parentheses behind the field name:
```javascript
<span className="editing-clone-attribute-name">
    {attr.fieldName}
    {!attr.visible ? " " + LocaleUtils.tr("editing.clone_hidden") : ""}
    {attr.autoCalculated ? " " + LocaleUtils.tr("editing.clone_calculated") : ""}
</span>
```

manisandro: Calculated fields (i.e. with `expression`) are marked as `readOnly` in the editConfig and won't be written by the data-service however [1]. For fields with a defaultValue it should work however.

[1] https://github.com/qwc-services/qwc-data-service/blob/master/src/dataset_features_provider.py#L1002

dominik-warch: The distinction between virtual fields (I assume this refers to calculated/expression fields) and fields with default values (which, as in my understanding, are also calculated) was not so clear to me in the QWC2 context. Yes, copying virtual, calculated fields does indeed make little sense.

manisandro: Yes virtual fields are expression fields (which are intrinsically read-only and do not exist in the data source).

manisandro:
So either the fields are missing from `targetFields`, are not mapped in `sourcePropertiesById`, or are of `type === 'list'` and hence discarded. Could you check?

dominik-warch:
There are no list items in my example. Based on my log statements, I think that matching only occurs when id and alias match. The sourceNameToId mapping is no longer working properly. Lines 346 and 347 should reference `field` and not `sourceFields` I think.

https://github.com/qgis/qwc2/blob/648535861bb48d18201a74a15b7c2ee82279ab67/plugins/Editing.jsx#L344-L348

Changed to

```javascript
[field.name]: field.id, 
[field.id]: field.id 
```

PR #524, opened 2025-11-07
add support for multipoint and -lines

rsrg-zwiama:
It would be nice to use the location recorder also for mulitpoints and -lines. Although it will stay a single geometry.


manisandro:
Thanks

PR #526, opened 2025-11-11
Visibility Presets Plugin: Save and restore user visibility presets without changing map position

NinaRoeckeleinSWM:
## User Story
Our users want the ability to save custom layer presets and quickly switch between them, without losing the current map position.

Relevant scenarios include:
- A user opens a link from another system to the QWC, which sets the correct map position. Now, they want to load their personalized layer configuration without loosing the location.
- A user frequently needs to review several specific combinations of layers for a given location as a part of their workflow.

## Implementation
- This feature is built using the existing bookmarks service, by omitting center, scale and extent when saving a bookmark.
- The original Bookmarks Plugin was generalized into a component to support both the standard Bookmarks plugin and the new LayerBookmarks plugin.

image</tab>&nbsp;
image

### Quick Select
- Additionally, an optional quick select menu is added to the BottomBar. Allowing users to switch between their saved presets with less clicks.
image

### Settings
- The LayerBookmark can also be selected as a default theme in the settings.
image

Looking forward to feedback!


danceb:
Great Idea! Just a few thoughts on this. 

As the functionality is very close to the existing bookmarks, I would prefer to extent the functionality of the bookmarks plugin instead of creating a new on with similar usability. 
An option could be to introduce a toggle switch to load the bookmark with or without center/extent and scale next to the other buttons within the bookmarks panel.

There is already an option to set a default map theme per user within the theme selection panel by clicking the star symbol at a specific map theme. How is this considered here? So which default configuration "wins" when opening the map viewer, the map theme or the bookmark?

So if there is a solition for this, I would recommend to set the default bookmark like the map theme with introducing the possibility to click a similar star symbol within the list entry of a bookmark (for instance on the right side). 
I think it is more understandable to set this within the plugin itself instead of within the settings planel.

manisandro:
Interesting functionality! As @danceb wrote, it would be interesting to integrate this into the existing bookmarks plugin. As far as I can see the core feature here is the quick switcher from the bottombar. What about the following behavior:

- When opening a bookmark from the bookmarks panel, or when set as default map, then the bookmark is restored with center / extent as is already the case
- When opening a bookmark fro the quick switcher, the center/extent of the bookmark to load is overridden with the current map extent
- Optionally, an additional button could be added to the bookmarks panel to open the bookmark while preserving the current extent

NinaRoeckeleinSWM:
@danceb Regarding the **selection of default themes/bookmarks**: This PR does not change how the defaults are managed. There can only be one default per user (regardless if it is a theme, bookmark or layerbookmark). Choosing a new default will override the old one. 

A **star icon** for setting defaults directly in the bookmark list would be a nice addition. We have dicussed the same idea here previously. We might implement this at some point in the future, but it is not within the scope of this PR.


@manisandro @danceb 
Regarding your suggestion to **integrate the new feature into the existing Bookmarks Plugin**:
We considered this before starting development. Let me explain the why we decided to create a separate plugin.
a) Although being similar from a technical perspective, bookmarks (with locations) and layer visibilities (without locations) are distinct features for our users, serving different purposes.
b) Our users would like to manage both as seperate lists (not mixed together)
c) the bookmarks plugin is already button heavy and our users don't find it so easy to navigate as is. Adding more complexity to it could make the user experience worse.
 
image

So yes, we are confident that keeping the two plugins seperate is the preferable solution (at the very least for our instance and our users).

dominik-warch:
I would like to support this suggestion. Predefined layer visibility sets are also something that our customer base uses.

In fact, it would be desirable to take this a step further and offer predefined selections that users can then expand themselves.

NinaRoeckeleinSWM:
Finally, we have updated the frontend to use the dedicated visibilitypresets endpoint in the permalink service. It took a while, because there were quiet a few changes to the Bookmarks Plugin (which this pull request builds upon). Now everything is cleanly build ontop of the latest changes. The functionality stays the same as initially described here. I have renamed everything to 'VisibilityPresets' consistently.

Also, one line needs to be added to the map viewer to enable loading visibility presets as a default visibility. I have added a tiny pull request on the map-viewer-repository.

components/StandardApp.jsx
```diff
@@ -158,11 +168,19 @@ class AppContainerComponent extends React.Component {
                 }
                 const initialTaskParam = (params.task ? JSON.parse(decodeURIComponent(params.task)) : null);
                 this.props.setCurrentTheme(theme, themes, false, initialExtent, layerParams, params.bl ?? null, state.layers, this.props.appConfig.themeLayerRestorer, this.props.appConfig.externalLayerRestorer, initialTaskParam);
+                if (state.visibilityPreset) {
+                    if (this.props.currentTheme) {
+                        this.props.setThemeLayersVisibilityPreset(state.visibilityPreset);
+                    } else {
+                        this.pendingVisibilityPreset = state.visibilityPreset;
+                    }
+                }
```
manisandro: I would recommend passing the initial visibility preset to `setCurrentTheme`.

NinaRoeckeleinSWM: Good Point. I have changed it.

manisandro:
Thanks!

manisandro:
Hmm testing a bit, there is one problem: restoring a visibility preset only makes sense in the theme it was created it. You'll probably need to store the theme name along with the preset configuration, and only show the visibility presets for the active theme in the panel.

NinaRoeckeleinSWM:
Yes, blindspot because we only have one theme in the setup used to develop this. Working on a solution to the problem tomorrow.

manisandro:
This is also causing problems in fullscreen 3D mode, the settings plugin is trying to read from the bookmarks stats which is not forwarded to the 3D store (see View3D.jsx), and hence the application crashes. I've reverted the commit for now.

NinaRoeckeleinSWM:
I have fixed the 3D issue on the branch. It doesn't show here since the PR is closed. I don't have rights to open it again. Would you do that, or should I open up a new one?

manisandro:
Please submit a new PR, merged PRs cannot be reopened.

PR #527, opened 2025-11-12
EditingSupport fix undefined geomType in .replace()

rsrg-zwiama:
in some rare cases, the 
`const geomType = this.props.editContext.geomType?.replace(/Z$/, '')`
fails, when geomType is none or undefined. It is the case, when 
1. open edit dialog
2. select dras
3. close edit dialog without any action
The problem also occures, when editing a relational Value (nested) of a not saved parent feature.

### DIFF
--- a/plugins/map/EditingSupport.jsx
+++ b/plugins/map/EditingSupport.jsx
@@ -100,7 +100,7 @@ class EditingSupport extends React.Component {
     addDrawInteraction = () => {
         this.reset();
         this.createLayer();
-        const geomType = this.props.editContext.geomType?.replace(/Z$/, '') || '';
+        const geomType = this.props.editContext.geomType.replace(/Z$/, '');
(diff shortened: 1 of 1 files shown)


plugins/map/EditingSupport.jsx
```diff
@@ -58,7 +58,7 @@ class EditingSupport extends React.Component {
     }
     render() {
         if (this.state.showRecordLocation) {
-            const geomType = this.props.editContext.geomType.replace(/Z$/, '');
+            const geomType = this.props.editContext.geomType?.replace(/Z$/, '') || '';
```
manisandro: What about
```
if (this.state.showRecordLocation && this.props.editContext.geomType) {
```
Using the `LocationRecorder` with an empty `geomType` will probably not work.

manisandro:
Thanks!

PR #539, opened 2025-12-17
fix(IdentifyUtils): get layer title value when feature info response …

benoitblanc:
…returns it

Hi,

I am not sure about all the logic when getting layername, layer.name or layer.title when parsing XML response from GetFeatureInfo request.

But I think there was an issue when `featureInfoReturnsLayerName` is false as it does not display layer title instead of layer name.

Thanks

### DIFF
--- a/utils/IdentifyUtils.js
+++ b/utils/IdentifyUtils.js
@@ -291,7 +291,7 @@ const IdentifyUtils = {
                 layername = layerEl.attributes.name.value;
                 layertitle = LayerUtils.searchSubLayer(layer, 'name', layername)?.title ?? layername;
             } else {
-                layertitle = layerEl.attributes.name.value;
+                layertitle = layerEl.attributes.title.value;
                 layername = LayerUtils.searchSubLayer(layer, 'title', layertitle)?.name ?? layertitle;
             }
 


manisandro:
The current code should be correct:
- `featureInfoReturnsLayerName=true` means that the `name` attribute of the `<Layer >`  element contains the technical layer name
- `featureInfoReturnsLayerName=false` means that the `name` attribute of the `<Layer >`  element contains the layer title

benoitblanc:
Well, my issue here is that I am using Mapproxy as an external Data URL for one of the layer in my project. For GetFeatureInfo results, I would like to get layer title instead of layer name.

image

Here is the response from Mapproxy GetFeatureInfo request:

```xml
<GetFeatureInfoResponse>
 <Layer title="My title" name="layer_name">
...
 </Layer>
</GetFeatureInfoResponse>
```

With this fix, I am able to get title correctly but it may break something as you are saying that the code is correct. So do you know how to handle this ?

Thanks

manisandro:
I see that the behavior of QGIS Server changed here [1], before it didn't write the `title` attribute, and furthermore I think QGIS Server 2 returned the layer title as `name` and QGIS Server 3 returns the layer name as `name`, which was the reason for introducing `featureInfoReturnsLayerName` back then.

But now discovering [1] is a good chance to clean this up. I've pushed [2] and [3], let me know if this works for you.

- [1] https://github.com/qgis/QGIS/commit/02cf7739b86e1fcce31a13b492ba7c986ab3ec53
- [2] https://github.com/qwc-services/qwc-feature-info-service/commit/06a98d8eb70ab98cd60bccfb605d8936a465cd32
- [3] https://github.com/qgis/qwc2/commit/ef5043fb5306abd183ea185f18dddf251db04158

benoitblanc:
Using `qwc-map-viewer-base:latest` with build from `qwc2/master` mounted in `/qwc2`, `qwc-feature-info-service:v2025.12.18` and `qwc-qgis-server:3.40`, it is working great !

Thanks a lot @manisandro

PR #569, opened 2026-02-18
Attribute form size nestedforms css hide

rsrg-zwiama:
alternative solution to https://github.com/qgis/qwc2/pull/519


rsrg-zwiama:
@manisandro 
I tried your idea with hiding the parent form via css. I tested this approach with a dataset of a parent features and two nested child features. It is possible to create new features with child and child's child feature without saving parent or first child before creating 2nd child. Also edits are cached and parent can be saved after the child is saved and vice versa.. I tested wich child features with and without geometry. And also with a nested child that has a not nested relation.
The editing window resizes to the content of child and parent features. For small screens the form switches to scroll. 
image

PR #579, opened 2026-02-27
Draft: QgisSearch: fix label when layer uses an expression as display name

### LINKED ISSUE #578: QGIS search displays empty results when QGIS layer defines an expression as Display Name
Hi,

The QGIS search displays features with empty labels when QGIS layer uses an expression as Display Name:

Image

Here's the expression I'm using in QGIS layer:

Image

The search query to the feature info service sets `with_maptip` to `false`, which leads to feature info service / QGIS Server not outputting the maptip attribute in the response.

# Fix suggestion

QgisSearch requests `GetFeatureInfo` with `WITH_MAPTIP` to `false` : https://github.com/qgis/qwc2/blob/0d4209f4b4ae6e450e454dba8a682b0770b8d97c/utils/SearchProviders.js#L236. It should be `true`.

tmuguet:
Fixes #578 

I now correctly have the results:

image

### DIFF
--- a/utils/SearchProviders.js
+++ b/utils/SearchProviders.js
@@ -233,7 +233,7 @@ class QgisSearch {
             HEIGHT: 100,
             LAYERS: [],
             FILTER: [],
-            WITH_MAPTIP: false,
+            WITH_MAPTIP: true,
             WITH_GEOMETRY: true,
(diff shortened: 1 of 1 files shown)


tmuguet:
Closing this one for now.

PR #606, opened 2026-03-23
feat: set startupTask through `task` URL parameter

benoitblanc:
Hi,

This PR allows to configure `startupTask` through URL parameters `stkey` and `stmode`. These parameters take precedence over per-theme configuration and global configuration parameter `startupTask`.

This could be used in external application which is interfacing QWC app with a configurable startup task.

I have a PR for documentation ready to push if this is accepted

Thanks for the review

### DIFF
--- a/components/StandardApp.jsx
+++ b/components/StandardApp.jsx
@@ -162,9 +162,8 @@ class AppContainerComponent extends React.Component {
             }
 
             const task = (params.task ? JSON.parse(decodeURIComponent(params.task)) : undefined)
-                || theme?.config?.startupTask
                 || ConfigUtils.getConfigProp("startupTask");
-            if (task) {
+            if (task && ((!params.task && !theme?.config?.startupTask) || (params.task && task !== theme?.config?.startupTask))) {
                 const mapClickAction = ConfigUtils.getPluginConfig(task.key).mapClickAction;
                 this.props.setCurrentTask(task.key, task.mode, mapClickAction, task.data);
             }
--- a/components/StandardApp.jsx
+++ b/components/StandardApp.jsx
@@ -156,16 +156,10 @@ class AppContainerComponent extends React.Component {
                 if (layerParams && ConfigUtils.getConfigProp("urlReverseLayerOrder")) {
                     layerParams.reverse();
                 }
-                this.props.setCurrentTheme(theme, themes, false, initialExtent, layerParams, params.bl ?? null, state.layers, this.props.appConfig.themeLayerRestorer, this.props.appConfig.externalLayerRestorer);
+                const initialTaskParam = (params.task ? JSON.parse(decodeURIComponent(params.task)) : null);
+                this.props.setCurrentTheme(theme, themes, false, initialExtent, layerParams, params.bl ?? null, state.layers, this.props.appConfig.themeLayerRestorer, this.props.appConfig.externalLayerRestorer, initialTaskParam);
             } else if (!ConfigUtils.havePlugin("Portal")) {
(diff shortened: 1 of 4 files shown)


benoitblanc:
I have made some modifications to handle new `task` URL parameter and to handle `taskData` in Editing plugin to be able to select layer and feature on componentDidUpdate

Tested with `task={"key":"Editing","data":{"layer":"qwc_demo#edit_polygons","feature":3}}` or `task={"key"="LayerTree"}` or `task={"key"="Measure","mode":"Polygon"}`

components/StandardApp.jsx
```diff
@@ -161,10 +161,12 @@ class AppContainerComponent extends React.Component {
                 this.props.showNotification("missingdefaulttheme", LocaleUtils.tr("app.missingdefaulttheme", params.t), NotificationType.WARN, true);
             }
 
-            const task = ConfigUtils.getConfigProp("startupTask");
-            if (task && !theme?.config?.startupTask) {
+            const task = (params.task ? JSON.parse(decodeURIComponent(params.task)) : undefined)
+                || theme?.config?.startupTask
```
manisandro: Per-theme startupTask handling occurs here: https://github.com/qgis/qwc2/blob/master/actions/theme.js#L121 - please keep the `if (task && !theme?.config?.startupTask) {` condition to avoid duplicate execution of `setCurrentTask`.

benoitblanc: I think that URL param `task` should take precedence on per-theme config `startupTask` and app config `startupTask` so I have adapted the condition

benoitblanc: @manisandro gentle ping :)

components/StandardApp.jsx
```diff
@@ -161,10 +161,11 @@ class AppContainerComponent extends React.Component {
                 this.props.showNotification("missingdefaulttheme", LocaleUtils.tr("app.missingdefaulttheme", params.t), NotificationType.WARN, true);
             }
 
-            const task = ConfigUtils.getConfigProp("startupTask");
-            if (task && !theme?.config?.startupTask) {
+            const task = (params.task ? JSON.parse(decodeURIComponent(params.task)) : undefined)
+                || ConfigUtils.getConfigProp("startupTask");
+            if (task && ((!params.task && !theme?.config?.startupTask) || (params.task && task !== theme?.config?.startupTask))) {
```
manisandro: This condition is quite hard to read. If I read it correctly, theme.config.startupTask will still override params.task, since `actions/theme.js@finishThemeSetup` will dispatch `setCurrentTask` if `theme.config.startupTask` is set, regardless of whether `params.task` is set.

I think it is easier to move the entire logic to `finishThemeSetup`, something like:
```
const taskParam = StandardApp.store.getState().localConfig.startupParams.task) ;
let task = null;
if (initialTheme && taskParam) {
    task = JSON.parse(taskParam);
} else if (theme.config.startupTask) {
    task = theme.config.startupTask;
} else if (initialTheme && ConfigUtils.getConfigProp("startupTask")) {
    task = ConfigUtils.getConfigProp("startupTask");
}
if (task) {
    const mapClickAction = ConfigUtils.getPluginConfig(task.key).mapClickAction;
    dispatch(setCurrentTask(task.key, task.mode, mapClickAction));
}
```
Or if `StandardApp.store.getState().localConfig.startupParams` is too ugly ev. pass it down as an optional `initialTask` param to `setCurrentTheme`.

benoitblanc: Done. There should be now only one call to `setCurrentTask`

manisandro:
Thanks! Would you like to also document the parameter here https://github.com/qwc-services/qwc-services.github.io/blob/master/src/topics/Interfacing.md ?

PR #614, opened 2026-04-08
add menuDisplayMode default value

jblanchg:
Hi!

Currently the Menu is always showing in the Portal plugin, that's because Portal does not pass menuDisplayMode to AppMenu, and AppMenu opens itself on mount when menuDisplayMode !== "normal", and this is always true as the prop is undefined. This should fix it.


manisandro:
Thanks!

PR #618, opened 2026-04-10
fix: bgLayer.params could be undefined

benoitblanc:
Hi,

When adding a WMS background layer with Themes plugin in QWC admin, there are no "params" and "version" added to the configuration so `bgLayer.params` is undefined.

Thanks


manisandro:
Thanks

PR #626, opened 2026-04-21
LayerTree: add visibleOnStartup option

dominik-warch:
Adds a `visibleOnStartup` prop to the `LayerTree` plugin that automatically opens the layer tree when a theme is loaded. The value can be set via the plugin `cfg` or via `theme.config.visibleOnStartup` and follows the existing `appMenuVisibleOnStartup` naming convention.

```json
  { "name": "LayerTree", "cfg": { "visibleOnStartup": true } }
```

This frees the startupTask for other options and also does not collide with the Portal plugin.

### DIFF
--- a/plugins/LayerTree.jsx
+++ b/plugins/LayerTree.jsx
@@ -83,8 +83,6 @@ class LayerTree extends React.Component {
         groupTogglesSublayers: PropTypes.bool,
         /** Whether to display the layer info button inside the layer settings menu rather than next to the layer title. */
         infoInSettings: PropTypes.bool,
-        /** Whether to open the LayerTree automatically when a theme is loaded. Can be set per-theme via the theme config. */
-        visibleOnStartup: PropTypes.bool,
         /** Default layer info window geometry with size, position and docking status. */
         layerInfoGeometry: PropTypes.shape({
             initialWidth: PropTypes.number,
@@ -134,6 +132,8 @@ class LayerTree extends React.Component {
         theme: PropTypes.object,
         toggleMapTips: PropTypes.func,
         transparencyIcon: PropTypes.bool,
+        /** Whether to open the LayerTree automatically when a theme is loaded. Can be set per-theme via the theme config. */
+        visibleOnStartup: PropTypes.bool,
         /** The initial width of the layertree, as a CSS width string. */
         width: PropTypes.string,
         zoomToExtent: PropTypes.func
--- a/plugins/LayerTree.jsx
+++ b/plugins/LayerTree.jsx
@@ -83,6 +83,8 @@ class LayerTree extends React.Component {
         groupTogglesSublayers: PropTypes.bool,
         /** Whether to display the layer info button inside the layer settings menu rather than next to the layer title. */
         infoInSettings: PropTypes.bool,
+        /** Whether to open the LayerTree automatically when a theme is loaded. Can be set per-theme via the theme config. */
+        visibleOnStartup: PropTypes.bool,
         /** Default layer info window geometry with size, position and docking status. */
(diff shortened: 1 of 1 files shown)


plugins/LayerTree.jsx
```diff
@@ -197,6 +199,12 @@ class LayerTree extends React.Component {
         if (this.props.layers !== prevProps.layers) {
             this.setState({activePreset: LayerUtils.getActiveVisibilityPreset(this.props.layers, this.props.theme.visibilityPresets)});
         }
+        if (this.props.theme !== prevProps.theme && this.props.theme?.id) {
+            const initiallyVisible = ConfigUtils.getConfigProp("visibleOnStartup", this.props.theme) ?? this.props.visibleOnStartup;
```
manisandro, line 203: If you specify the per-theme plugin config prop via
```
"config": {
  "plugins": {
    "LayerTree": {"visibleOnStartup": true}
  }
}
```
it will get merged automatically into the props passed to the component. No need to use `ConfigUtils.getConfigProp`.

plugins/LayerTree.jsx
```diff
@@ -197,6 +199,12 @@ class LayerTree extends React.Component {
...
+            if (initiallyVisible) {
+                this.props.setCurrentTask("LayerTree");
```
manisandro, line 205: Does this really work in parallel with `startupTask: <some other task>`? Won't one end up cancelling the other? If it works, I suspect it is rather by accident, depending on the component initialization order.

dominik-warch: It depends on the startupTask, yes! My primary use case for this is not using it as startupTask so it does not conflict with the Portal plugin which only shows for a split second before the startupTask kicks in.

manisandro: The Portal per se shows only if there is no `t=` in the URL. Am I guessing right that you'd like the portal plugin not to show initially, but only when explicitly invoked from say the app menu? In that case a config prop for the Portal Plugin "showOnStartup: false" or similar could be a more targeted solution.

dominik-warch: No, I want the Portal to show on startup. But the startupTask opening the LayerTree closes the Portal immediatly. Without selecting a theme in the Portal and without a defaultTheme (so no `t=` in the URL) I end up with an empty canvas and an empty opened LayerTree.

manisandro: What version of qwc are using? If you are using earlier than `v2026.04.19`, can you try upgrading? The commit https://github.com/qgis/qwc2/commit/bd5e4b5f5ceeb2325fb25f1ef89d184b30040cab might in fact fix this issue.

dominik-warch: The newest version seems even weirder: I do not see the Portal at initial load (not even for a split second) and there is no LayerTree started, only the blank canvas (because I have no defaultTheme). If I switch themes via ThemeSwitcher, the LayerTree opens automatically. If I switch themes via Portal plugin after the initial load, the LayerTree does not open.

manisandro: Odd, I cannot reproduce this. If my themes.json has `defaultTheme: null`, I see the Portal, regardless of whether a `startupTask` is set or not. When I select a theme and have a  `startupTask`, that task is activated when switching to the first theme.

dominik-warch: After the latest update, it's working for me now, too. I'm not sure why it took so long.

Is the PR still a good addition, or should I close it?

manisandro: I'd close it, the regular startupTask covers this.

PR #630, opened 2026-04-24
Add translation for IdentifyRadius menu

benoitblanc:
This adds translation when using Identify plugin with Radius mode as menu.

Thanks


manisandro:
Thanks!

PR #631, opened 2026-04-24
EditComboField: add support for updating values from props

tmuguet:
Currently, updating the `values` props of EditComboField is not taken into account.
This enables refreshing the state from the props if the **reference** is updated.
For simplicity, I chose not to handle the logic of checking if `value` is still among the new values, it's up to the parent component to do so.


manisandro:
Thanks

PR #637, opened 2026-04-29
fix(MapFilter): state may not be initialized when rendering predefined filters

benoitblanc:
Hi,

When using MapFilter plugin as `startupTask` in a theme, state is not initialized before rendering predefined filters. This PR aims to fix this issue (`can't access property "values", n.state.filters[e.id] is undefined`).

This is testable with qwc_demo project.

Is it something that could be reported to `lts` branch ?

Thanks

### DIFF
--- a/plugins/MapFilter.jsx
+++ b/plugins/MapFilter.jsx
@@ -431,12 +431,12 @@ class MapFilter extends React.Component {
     };
     renderPredefinedFilters = () => {
         const predefinedFilters = this.collectPredefinedFilters(this.props.layers);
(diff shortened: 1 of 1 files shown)


manisandro:
Perhaps cleaner to omit rendering the predefined filter entirely in `renderPredefinedFilters` if the `config.id` does not exist in `this.state.filters`?

benoitblanc:
I have applied a filter to the predefined filters before to render them.

manisandro:
Thanks!

PR #640, opened 2026-04-30
fix(theme):  also get startupTask prop from per-theme mobile or desktop section

benoitblanc:
Hi,

I would like to override `startupTask` per-theme, but also only for "desktop" mode.

This PR allows it by getting prop using ConfigUtils.getConfigProp when there is no initialTask.

```js
getConfigProp(prop, theme, defval = undefined) {
        const section = config.isMobile ? "mobile" : "desktop";
        const themeProp = theme?.config?.[section]?.[prop] ?? theme?.config?.[prop];
        return themeProp ?? config[section]?.[prop] ?? config[prop] ?? defval;
}
```

Before, I was only checking for `theme?.config?.startupTask` for per-theme config, which is not enough for this use-case.

Thanks

### DIFF
--- a/actions/theme.js
+++ b/actions/theme.js
@@ -119,7 +119,8 @@ export function finishThemeSetup(dispatch, theme, themes, layerConfigs, insertPo
         switching: false
     });
 
-    const task = initialTask || theme?.config?.startupTask || (initialTheme ? ConfigUtils.getConfigProp("startupTask") : null);
+    const section = ConfigUtils.isMobile() ? "mobile" : "desktop";
+    const task = initialTask || (theme?.config?.[section]?.startupTask ?? theme?.config?.startupTask) || (initialTheme ? ConfigUtils.getConfigProp("startupTask") : null);
     if (task) {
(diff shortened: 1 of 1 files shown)


actions/theme.js
```diff
@@ -119,7 +119,7 @@ export function finishThemeSetup(dispatch, theme, themes, layerConfigs, insertPo
         switching: false
     });
 
-    const task = initialTask || theme?.config?.startupTask || (initialTheme ? ConfigUtils.getConfigProp("startupTask") : null);
+    const task = initialTask || ConfigUtils.getConfigProp("startupTask", theme);
```
manisandro: This will cause the startupTask at toplevel config.json to be fired every time a theme is switched, which is different than the current behaviour (the startupTask from toplevel config.json is only activated when the viewer is initially loaded)

benoitblanc: Thanks, I don't think I was aware about this behavior.

I have updated by getting only the relevant part of `getConfigProp` (`const themeProp = theme?.config?.[section]?.[prop] ?? theme?.config?.[prop];`)

manisandro:
Thanks!

PR #643, opened 2026-05-05
Fix Cyclomedia Plugin for many points

BenediktSeidlSWM:
Commit 0982ee9 already went into the right direction but missed two issues:

- There may be more than 3000 features even in a quarter of the extent.
- The WFS service sometimes does not return features on the border of an extent (due to rounding issues).

This pull request calculates the number of tiles based on the actual tile size obtained with a heuristic and adds an overlap between the tiles.


manisandro:
Thanks!

PR #653, opened 2026-05-27
MapExport: document that allowedFormats controls sort order and list supported mimetypes

dominik-warch:
While checking whether MapExport supports ordering export formats like the Identify plugin's `enableExports`, I found that it already does. Updated the JSDoc comment and `doc/plugins.md` to make this explicit, including the list of supported mimetypes.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -739,7 +739,7 @@ Allows exporting a selected portion of the map to a variety of formats.
 
 | Property | Type | Description | Default value |
 |----------|------|-------------|---------------|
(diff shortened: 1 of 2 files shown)


manisandro:
Hi @dominik-warch 

Thanks, can you put the full docstring in MapExport.jsx and then run yarn run plugindoc to regenerate plugins.md?

dominik-warch:
Ah sure! I updated the docstring in MapExport.jsx and regenerated plugins.md via yarn run plugindoc. The output is identical to the manual edit from the first commit because I used the same wording, so no change in the commit.

manisandro:
Thanks!

PR #655, opened 2026-05-27
LocateSupport: add defaultFollowScale prop to zoom to position on initial locate

dominik-warch:
My proposed solution based on issue #607

### DIFF
--- a/plugins/map/LocateSupport.jsx
+++ b/plugins/map/LocateSupport.jsx
@@ -29,10 +29,10 @@ class LocateSupport extends React.Component {
     static propTypes = {
         changeLocatePosition: PropTypes.func,
         changeLocateState: PropTypes.func,
-        /** Scale denominator to zoom to when the position is first determined. If not set, the map zoom level is not changed. */
-        defaultFollowScale: PropTypes.number,
         /** Whether to draw an accuracy circle around the location point. */
         drawCircle: PropTypes.bool,
+        /** Scale denominator to zoom to when entering follow mode. If not set, the map zoom level is not changed. */
+        followScale: PropTypes.number,
         locateState: PropTypes.object,
         map: PropTypes.object,
         /** Whether to display the accuracy in meters (`true`) or in feet (`false`). */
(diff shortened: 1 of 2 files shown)


plugins/map/LocateSupport.jsx
```diff
@@ -27,6 +29,8 @@ class LocateSupport extends React.Component {
     static propTypes = {
         changeLocatePosition: PropTypes.func,
         changeLocateState: PropTypes.func,
+        /** Scale denominator to zoom to when the position is first determined. If not set, the map zoom level is not changed. */
+        defaultFollowScale: PropTypes.number,
```
manisandro: I think this name is misleading? The logic below applies this default scale whenever the position is first determined, not when the user switches to FOLLOW mode. Which behavior is desired? I think it actually makes sense to apply the scale only on FOLLOW - if this is the goal, the logic below would need to be changed, instead you could react on a state change in `componentDidUpdate`.

dominik-warch: > I think this name is misleading? The logic below applies this default scale whenever the position is first determined, not when the user switches to FOLLOW mode. Which behavior is desired? I think it actually makes sense to apply the scale only on FOLLOW - if this is the goal, the logic below would need to be changed, instead you could react on a state change in `componentDidUpdate`.

Yes. I moved the the logic to `componentDidUpdate` to apply on every transition to `FOLLOWING`. I also renamed the prop to `followScale` since it does not really represent a fallback value.

plugins/map/LocateSupport.jsx
```diff
@@ -110,6 +116,13 @@ class LocateSupport extends React.Component {
                 this.requestedMode = "DISABLED";
                 this.stop();
             }
+            if (newState === "FOLLOWING" && oldState !== "FOLLOWING" && this.props.followScale && this.props.scales) {
+                const mapPos = this.geolocate.getPosition();
+                if (mapPos) {
+                    const wgsPos = CoordinatesUtils.reproject(mapPos, this.props.projection, "EPSG:4326");
```
manisandro: This is unnecessary. You can pass the `mapPos` to `zoomToPoint` directly, passing `this.props.projection` as the CRS.

dominik-warch: Fixed it

manisandro:
Thanks!

PR #656, opened 2026-05-27
2D plugin functionality also in 3D plugins

danceb:
This PR adds the following functions from the 2D bottom bar also to the 3D version in a first attempt:
- `additionalBottomBarLinks`
- `displayCoordinates`
- `termsUrl`
- `termsUrlIcon`
- `termsUrlTarget`
- `viewertitleUrl`
- `viewertitleUrlIcon`
- `viewertitleUrlTarget`

possible improvements could be:
- using `additionalBottomBarLinks`, `termsUrl*` and `viewertitleUrl*` from the BottomBar config instead of extra duplicated parameters (is that possible?)
- don't display `additionalBottomBarLinks`,  `termsUrl*` and `viewertitleUrl*` in splitscreen mode, as they are already visible in the 2D BottomBar, like it is done with the 2D menu and topbar items

Any ideas about the last two things? Thanks for review!

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -1331,6 +1331,10 @@ Map button for switching the background layer of the 3D map.
 
 Bottom bar of the 3D map, displaying coordinates, projection, etc.
 
+| Property | Type | Description | Default value |
+|----------|------|-------------|---------------|
+| displayCoordinates | `bool` | Whether to display the coordinates in the bottom bar. | `true` |
+
 ## Compare3D<a name="compare3d"></a>
 
 Split-screen and compare objects in the 3D map.
@@ -1364,6 +1368,8 @@ Layer and object tree for the 3D map
 |----------|------|-------------|---------------|
 | groupTogglesSublayers | `bool` | Whether toggling a group also toggles all sublayers. | `undefined` |
 | importedTilesBaseUrl | `string` | Base URL of imported tile sets. | `':/'` |
+| side | `string` | The side of the application on which to display the sidebar. | `'right'` |
+| width | `string` | The initial width of the layertree, as a CSS width string. | `'25em'` |
 
 ## MapCopyright3D<a name="mapcopyright3d"></a>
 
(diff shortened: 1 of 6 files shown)


manisandro:
@danceb Thanks! You could get the configuration of the 2D BottomBar via something like
```
const config = ConfigUtils.getPluginConfig("BottomBar")?.cfg ?? {};
```
to avoid duplication. I think it makes sense to not display the links in split screen mode.

danceb:
@manisandro is there a way to not use `viewMode` from `props`, as there is no use to define this on user site? For instance via `state`?

manisandro:
Can you elaborate on your question? Not sure I understand

danceb:
To achive the case, that the links are not shown in split view mode, I assigned `viewMode` within `propTypes`:
```
    static propTypes = {
        [...]
        viewMode: PropTypes.number
    };
```
and use this for filter the elements:
```
{this.props.viewMode === ViewMode._3DFullscreen ? (
    <span className="bottombar-links">
        {leftBottomLinks}
    </span>
) : null}
```
But with this, `viewMode` became a parameter, which could be set in the plugin config, or am I wrong? I think, this should not happen.

manisandro:
No, props are passed to the component either from a parent component (this is how the config.json settings are passed to the plugin), or via `connect`, which extracts parts of the global application state and passes them to the component as props. So here
```
export default connect(() => ({}), {
export default connect((state) => ({
    fullscreen: state.display?.fullscreen,Expand commentComment on line R140Fixed
    viewMode: state.display?.viewMode
}), {
    openExternalUrl: openExternalUrl,
    setBottombarHeight: setBottombarHeight
})(BottomBar3D);
 ```
`viewMode` is read from the global application `state.display.viewMode`.

manisandro:
Can you run `yarn run plugindoc` to update the plugin docs?

manisandro:
Thanks!

PR #662, opened 2026-06-10
Disable scaling in MapExport when no frame is given

BenediktSeidlSWM:
The missing parameter leads to weird glitches in the MapExport plugin, since 8c6a0c0ae8bde25c2afed3fb8be98679e37b53a2

### DIFF
--- a/plugins/MapExport.jsx
+++ b/plugins/MapExport.jsx
@@ -276,7 +276,7 @@ class MapExport extends React.Component {
                 center={this.props.map.center} fixedFrame={frame} geometryChanged={this.geometryChanged} key="PrintSelection" scale={this.state.scale}
             />);
         } else {
-            return (<PrintSelection allowRotation={false} geometryChanged={this.geometryChanged} key="PrintSelection" />);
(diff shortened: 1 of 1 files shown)


manisandro:
What weird glitches? As far as I can see, it works correctly and it is useful to be able to resize an existing selection?

BenediktSeidlSWM:
The glitch we noticed is hard to reproduce, but if the user selects a box and then quickly tries to select another box after releasing the mouse, the map moves instead of drawing another rectangle. We were able to avoid that issue by disabling the resizing. Also I think it is confusing that the user cannot move the extent nor change the aspect ratio after selecting a box but only scale it around the intial center. In my opinion, it is easier to just select a new rectangle.

BenediktSeidlSWM:
We could also make this an option by defining `allowScaling={this.props.allowedScales !== false}` like it is defined a few lines above.

manisandro:
Uhm so far haven't been able to reproduce the glitch. Regarding `allowScaling={this.props.allowedScales !== false}`, in non-fixed-page-mode, the size of the selection is actually not related to the map scale, changing the map scale will just result in a larger/smaller image being requested for the selected area.

In my view resizing the rectangle can actually be useful, though more useful if were actually a free resize, not aspect-ratio preserving scaling.

PR #663, opened 2026-06-10
Visibility Presets Plugin

NinaRoeckeleinSWM:
Continuation of https://github.com/qgis/qwc2/pull/526/

**Changes:**
- Fix crashes in 3D mode
- Visibility presets are now filtered to the currently active theme
- Bookmarks now save the `theme_id` explicitly too. But no behavioral change to in the Bookmark plugin.
- The quick-select dropdown filters both bookmarks and visibility presets by the currently active theme


plugins/Portal.jsx
```diff
@@ -70,7 +70,7 @@ class Portal extends React.Component {
         menuVisible: false
     };
     componentDidMount() {
-        if (!this.props.startupParams.t && !this.props.startupParams.k && !this.props.startupParams.bk) {
+        if (!this.props.startupParams.t && !this.props.startupParams.k && !this.props.startupParams.bk && !this.props.startupParams.vp) {
             this.props.setCurrentTask("Portal");
```
manisandro, line 74: As far as I see `vp` will require `t` in the URL anyway? So if `t` is not set, `vp` should be ignored?

plugins/Settings.jsx
```diff
@@ -131,7 +131,12 @@ class Settings extends React.Component {
         }
         const themeNames = ThemeUtils.getThemeNames(this.props.themes);
         const bookmarks = (this.props.bookmarks || []);
-        const visibilityPresets = (this.props.visibilityPresets || []);
+        const defaultThemeId = this.props.themes?.defaultTheme;
+        const visibilityPresets = (this.props.visibilityPresets || []).filter(vp =>
```
NinaRoeckeleinSWM: Thinking about your comment. I found that the way I had added them to default theme settings, wouldn't work for multi-theme instances. Now only displaying options for the default theme.

plugins/Portal.jsx
NinaRoeckeleinSWM: This is true. I changed it.

manisandro: Actually, since the preset data also stores the theme id, the theme could be determined when restoring the preset, allowing starting the viewer with only the `vp` parameter?

NinaRoeckeleinSWM: I have implemented loading the theme from the vp-parameter. Still needs small backend change. This assumes that
`/api/permalink/visibility_presets/{key}` returns
```
{
    "preset": {Preset Object},
    "theme_id": "Some ID"
}
```.

NinaRoeckeleinSWM:
Opening the viewer with only the vp-parameter is now possible. The theme is loaded as saved in the preset.
Thank you @manisandro for the backend work!

manisandro:
Thanks! I've merged it and did some testing, looks good so far.

NinaRoeckeleinSWM:
Thank you. I'm glad we can close it. Hope it's all fine now. If something comes up, get in touch.

PR #664, opened 2026-06-10
Add lockedWindow support to AttributeTable

HermanAssink:
The existing lockedWindow implementation used by MapLegend cannot be applied directly to AttributeTable because AttributeTable depends on task activation and task data (state.task.data.layer). A direct copy causes the table to open at startup and can prevent reopening after a detached window is closed. This implementation preserves normal task activation while keeping the table visible after activation.

Behavior:
- AttributeTable opens on task activation.
- AttributeTable remains visible when switching to other tasks.
- AttributeTable can be reopened after a detached window is closed.
- Existing behavior is unchanged when lockedWindow is false.

Use case: When editing attribute data it is practical to keep AttributeTable open and preserve state while being able to use LayerTree, Identify or other taks.

### DIFF
--- a/plugins/AttributeTable.jsx
+++ b/plugins/AttributeTable.jsx
@@ -5,18 +5,16 @@
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
  */
-
 import React from 'react';
 import {connect} from 'react-redux';
-
 import PropTypes from 'prop-types';
 
-import {setCurrentTask} from '../actions/task';
-import AttributeTableWidget from '../components/AttributeTableWidget';
-import ResizeableWindow from '../components/ResizeableWindow';
-import EditingInterface from '../utils/EditingInterface';
-import LocaleUtils from '../utils/LocaleUtils';
+import {setCurrentTask} from 'qwc2/actions/task';
 
+import AttributeTableWidget from 'qwc2/components/AttributeTableWidget';
+import ResizeableWindow from 'qwc2/components/ResizeableWindow';
+import EditingInterface from 'qwc2/utils/EditingInterface';
+import LocaleUtils from 'qwc2/utils/LocaleUtils';
 
 /**
  * Display the attribute table of layers in a dialog.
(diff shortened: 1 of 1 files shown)


plugins/AttributeTable.jsx
```diff
@@ -5,18 +5,16 @@
...
-import AttributeTableWidget from '../components/AttributeTableWidget';
-import ResizeableWindow from '../components/ResizeableWindow';
-import EditingInterface from '../utils/EditingInterface';
-import LocaleUtils from '../utils/LocaleUtils';
+import {setCurrentTask} from 'qwc2/actions/task';
 
+import AttributeTableWidget from 'qwc2/components/AttributeTableWidget';
+import ResizeableWindow from 'qwc2/components/ResizeableWindow';
+import EditingInterface from 'qwc2/utils/EditingInterface';
+import LocaleUtils from 'qwc2/utils/LocaleUtils';
```
manisandro, line 17: I guess these changes are not necessary?

manisandro: Thanks for the PR. One potential issue I see here is that if you open a dataset in the Attribute Table, then, while the attribute table is open, edit a feature of that dataset with the Edit panel, the Attribute Table will not be refreshed and potentially show an old version of the feature. If you later perform a change in the Attribute Table to that feature, you might end up overwriting the changes performed via the Edit panel.

HermanAssink: Hi Sandro Mani, Thanks for your quick response. Indeed, "qwc2" in the import paths must remain "..".
Regarding the risk of conflicting edits I propose the following solution: A Redux-based edit-lock design. Namely:

1. Add actions/editingLock.js
2. Add reducers/editingLock.js
3. Add editingLock to reducers/index.js
4. In FeatureForm.jsx and AttributeTableWidget.jsx set editingLock when editing starts and release the lock when editing ends.

Is this type of solution acceptable?

Regards, Herman Assink

manisandro: I agree with the lock idea. You can add the locking actions/reducers directly to the existing `actions/editing` and `reducers/editing` files.

manisandro: Also, you can handle locking directy in the AttributeForm component, so you'll cover `Editing` and `FeatureForm` in one go.

PR #667, opened 2026-06-17
Update updateTranslations - typescript, custom path allowed

Marshlyin:
This PR improves the translation mechanism for custom projects built on qwc2 that wouldn't fit the initial architecture of qwc2. 
For example, this allows the use of typescript components (tsx) with the update of the files regex. 
`const files = listDir(topdir, /^.*\.[jt]sx?$/);`
Also, it allows applications with an "src" architecture to work with the --basePath="..." attribute  (that defaults to /js if not specified) of the tsUpdate node script.


manisandro:
LGTM, thanks

PR #668, opened 2026-06-17
impr: add named substitution to the translation mecanism

Marshlyin:
This PR improves the translation mechanism and adds support for key based translations. 
It allows the user to use the LocaleUtils.tr() function with an object as second parameter, and to use named keys in the json translation file. The default mechanism is this available. 
For example, the translation : 
`"test": "this is a {foo} example. this is another {bar} example."`
Can be used with : 
`LocaleUtils.tr("test", {foo: "good", bar: "nice"})`

### DIFF
--- a/utils/LocaleUtils.js
+++ b/utils/LocaleUtils.js
@@ -75,12 +75,16 @@ const LocaleUtils = {
             }
         });
     },
-    tr(key) {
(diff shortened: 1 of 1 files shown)


manisandro:
Thanks!

PR #674, opened 2026-06-18
Update updateTranslation.js regex to support kebab case

Marshlyin:
This PR adds the possibility for the user to use kebab-case for the translation keys, as well as already supported camelCase, snake_case or PascalCase.


manisandro:
:+1:

PR #678, opened 2026-06-25
Allow loading the news html instead of embedding an iframe

BenediktSeidlSWM:
Instead of loading the news document in an iframe, this patch allows the plugin to load the html via http and embed it directly in the document. The mechanism works exactly as in the Help plugin (in fact most of the code is copied from there). The default is still to embed an iframe.


manisandro:
Thanks

PR #682, opened 2026-07-02
Rework ServiceInfo as a plugin

benoitblanc:
Hi,

This PR aims to be able to use ServiceInfo as a plugin. We don't need `setActiveServiceInfo` action anymore.

We can now add the following configuration for the plugin in `config.json`:

```json
      {
        "name": "ServiceInfo",
        "cfg": {
          "showLayerTree": true # default is false
        }
      },
```

The behavior in LayerTree is still the same thanks to (new) components `ServiceInfoWindow` and `ServiceInfoWidget`.

_This is funded by "Les Agences de l'eau | Direction des Systèmes d’Information et des Usages Numériques (https://www.lesagencesdeleau.fr/ )"_

cc @naub1n

Thanks for the review


manisandro:
This adds quite a lot of duplicate code. Why can't you just keep the component/ServiceInfoWindow more-or-less as is, and if you want a task controller ServiceInfoWindowPlugin.jsx, just render the <ServiceInfoWindow> there, passing the props as necessary?

Also, the "showLayerTree" option makes me think whether it is time to rethink "modal" vs "non-modal" tasks. Currently we have the cases where:
- The global redux task state controls whether a plugin component is active
- The global redux task only triggers an internal component state to change to visible, and then resets, but while doing so clearing the previous task.

The second case could be more elegantly be covered by say a `start/stopNonModalTask(<id>)` actions, which control a `nonmodalTasks` list in the redux state without altering the modal task. The ServiceInfoWindow could then become a non-modal task, perhaps declared with a `nonModalTask = true` static class field similar to `availableIn3D`. If a task is then triggered via menuItem/toolbarItem/taskButton, the task would be started modal/nonmodal based on this flag.

This way, a user could open the ServiceInfoWindowPlugin with an open LayerTree and not lose the LayerTree doing so.

What do you think?

benoitblanc:
> Why can't you just keep the component/ServiceInfoWindow more-or-less as is, and if you want a task controller ServiceInfoWindowPlugin.jsx, just render the there, passing the props as necessary?

Do you mean to have plugin just rendering ServiceInfoWindow component instead of ResizeableWindow+ServiceInfoWidget ? I think that I've forgot to do this part :thinking: ... :tired_face:

For modal/non-modal tasks, indeed, it needs some work to enhance this behavior. I don't have strong opinion about your solution. I need some more time to think about it. For now, do I keep this `showLayerTree` option until work on tasks is done ?

Thanks

manisandro:
I guess we can look at the non-modal task aspect in a second moment

manisandro:
Sorry but what is the point to split off ServiceInfoWidget from ServiceInfoWindow, just to have ServiceInfoWindow render ServiceInfoWidget as the only consumer of that subcomponent? I see no use of a separate ServiceInfoWidget?

benoitblanc:
> Sorry but what is the point to split off ServiceInfoWidget from ServiceInfoWindow, just to have ServiceInfoWindow render ServiceInfoWidget as the only consumer of that subcomponent? I see no use of a separate ServiceInfoWidget?

Sorry, It was for the first attempt of the PR but it does not make sense anymore.

manisandro:
Thanks

PR #683, opened 2026-07-02
Show popup info when hovering theme item in list

benoitblanc:
Hi,

This PR aims to show a popup that can get HTML content from project abstract.

Before (no HTML content interpreted):

image

After:

image

It also fixes some style on ThemeList when item has a description.

Before:

image

After:

image

_This is funded by "Les Agences de l'eau | Direction des Systèmes d’Information et des Usages Numériques (https://www.lesagencesdeleau.fr/)"_

cc @naub1n

Thanks for the review

### DIFF
--- a/components/ThemeList.jsx
+++ b/components/ThemeList.jsx
@@ -7,6 +7,7 @@
  */
 
 import React from 'react';
+import ReactDOM from 'react-dom';
 import {connect} from 'react-redux';
 
 import axios from 'axios';
@@ -51,8 +52,13 @@ class ThemeList extends React.Component {
     };
     state = {
         expandedGroups: {},
+        hoveredItemId: null,
         visibleThemeInfoMenu: null
     };
+    constructor(props) {
+        super(props);
+        this.el = {};
+    }
     groupMatchesFilter = (group, filter) => {
         if (group && group.items) {
             for (let i = 0, n = group.items.length; i < n; ++i) {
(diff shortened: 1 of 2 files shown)


components/ThemeList.jsx
```diff
@@ -123,22 +129,38 @@ class ThemeList extends React.Component {
...
                             key={item.id}
                             onClick={() => this.setTheme(item)}
-                            title={title}
+                            onMouseEnter={() => this.setState({hoveredItemId: item.id})}
+                            onMouseLeave={() => this.setState({hoveredItemId: null})}
+                            ref={(el) => {this.el[item.id] = el; }}
                         >
-                            <div className="theme-item-title" title={item.title}>
+                            {title && isCurrentItemHovered && this.el[item.id] && (
+                                ReactDOM.createPortal(
```
manisandro, line 152: Maybe I'm overlooking something, but why are you creating a portal to render the fragment in a direct descendant of the parent element? Rendering the element without a portal should give you the same result. OTOH a portal makes sense to render outside the container, so i.e. to break out of the scroll viewport.

manisandro:
@benoitblanc Are you still working on this?

benoitblanc:
Yes but I have some days off. Will come back in September and I will rework on this PR

benoitblanc:
Hi @manisandro 

Indeed, I have deleted the React portal as it was not needed. I have done some tests to show the popup on the mouse cursor instead of the theme item bottom. Not sure what is the best for the user...

components/ThemeList.jsx
```diff
@@ -122,22 +127,36 @@ class ThemeList extends React.Component {
...
                         title += LocaleUtils.tr("themeswitcher.match.keywords") + ": " + item.keywords;
                     }
+                    const isCurrentItemHovered = this.state.hoveredItemId === item.id;
                     return (
                         <li className={activeThemeId === item.id ? "theme-item theme-item-active" : "theme-item"}
                             key={item.id}
                             onClick={() => this.setTheme(item)}
-                            title={title}
+                            onMouseEnter={() => this.setState({hoveredItemId: item.id})}
+                            onMouseLeave={() => this.setState({hoveredItemId: null})}
```
manisandro, line 146: Can this be solved purely via CSS as done in the `MenuButton` component, see https://github.com/qgis/qwc2/blob/master/components/widgets/style/MenuButton.css#L85?

benoitblanc: Yes, I tried with something like (and no JS events) : 

```css
div.ThemeList div.theme-item-tooltip {
    position: absolute;
    display: none;
    width: max-content;
    height: auto;
    left: 50%;
    top: 100%;
    transform: translateX(-50%);
    min-width: 100%;
    max-width: 150%;
    max-height: calc(50vh - var(--topbar-height) - var(--bottombar-height));

    background-color: var(--tooltip-bg-color);
    border: 1px solid var(--tooltip-border-color);
    color: var(--tooltip-text-color);
    font-weight: normal;
    padding: 0.5em;
    border-radius: 0.5em;
    white-space: normal;
    overflow: auto;
    z-index: 3;
}

div.ThemeList li:hover:not(:has(div.theme-item-icons:hover)) div.theme-item-tooltip {
    display: inline;
}

.sidebar-body:has(.theme-item:hover) {
    overflow: visible !important;
}
```

But I didn't manage to reduce width when item is right to the border so the popup is outside the screen.
Do you think it is a best way to show the popup ?

benoitblanc: @manisandro what do you think about this solution ? ^

image

manisandro: Sorry haven't yet found time to look into this in detail, I'll try to do so by the end start of next week. (If a pure CSS solution works I'd prefer it rather than using the state variable which triggers a rerender).

PR #684, opened 2026-07-07
feat: params.featuresearch to exclude search provider from ObjectSearch

dominik-warch:
Single-field qgis providers meant only for the TopBar quick search were also showing up in the ObjectSearch (FeatureSearch) sidebar, since it lists any qgis provider with params. This PR adds an opt-out flag, `params.featuresearch: false`, to exclude a provider from that dialog explicitly.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -486,6 +486,11 @@ Displays a dialog with a search form for configured QGIS feature searches with o
 
 See [Configuring the QGIS feature search](../../topics/Search/#configuring-the-qgis-feature-search).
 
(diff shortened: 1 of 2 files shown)


manisandro:
Thanks - would you mind moving the documentation to this page [1]? (aka [2])

- [1] https://docs.qwc.app/master/topics/Search/#configuring-the-qgis-feature-search
- [2] https://github.com/qwc-services/qwc-services.github.io/blob/master/src/topics/Search.md

dominik-warch:
Sure! I removed the documentation and will create a separate PR in the other repo!

manisandro:
Thanks!

Issue #677, opened 2026-06-25
PDF print in 3D View


danceb:
If the print section is greater than the map window (which seems to be the default), like:

<img width="769" height="472" alt="Image" src="https://github.com/user-attachments/assets/250cc29d-1bda-4a33-96b5-5ba3d46e4eb1" />

the print results looks something like this:

<img width="492" height="684" alt="Image" src="https://github.com/user-attachments/assets/af185431-354b-408e-bdec-d4d100179d3f" />

I think the fix would be to not allow a greater print section than the extent of the map window, as this works as expected:

<img width="390" height="474" alt="Image" src="https://github.com/user-attachments/assets/2302899e-3431-495c-a191-8a4c8d0bcc35" />

<img width="329" height="461" alt="Image" src="https://github.com/user-attachments/assets/e72e8dab-d482-46d5-bc92-dcd1415d1196" />

manisandro:
Should be less easy to draw an out-of-screen selection with [1] (and [2] also helps with initial positioning).


- [1] https://github.com/qgis/qwc2/commit/852e440920fd35541826af596a3151d669d0fb3f
- [2] https://github.com/qgis/qwc2/commit/b901beb5ff832d45bfc593b68cf9ea9eacaa577e

danceb:
Unfortunately the issue is not solved ...

If PDF is selected as output format, the printing box is locked now at top and bottom and can be moved sidewise only. Chaning the size of box by dragging the edges is not possible anymore. This state stays, if I am selection another file format afterwards.

manisandro:
Further improved with https://github.com/qgis/qwc2/commit/ae7144b94e164b5acf1f1bd86d040a50cc3e6777

PR #693, opened 2026-07-10
MapExport3D improvements

danceb:
This PR improves the changes made related to https://github.com/qgis/qwc2/issues/677 a little bit more.

For PDF export the selection consider 75 % percent of the window width and height, and for landscape print templates, the selection is based on the width and not the height.

Furthermore, the export formats of the plugin could be set with a config parameter.


manisandro:
Thanks, could you also run `yarn run plugindoc` to update the docs with the new parameter?

manisandro:
Thanks!

PR #696, opened 2026-07-13
Fix defaultBackgroundLayers fallback for empty backgroundLayers

stephaschu:
Empty backgroundLayers arrays prevent the defaultBackgroundLayers fallback and can be set by qwc-admin-gui or qwc-map-viewer: 
https://github.com/qwc-services/qwc-admin-gui/blob/master/src/plugins/themes/controllers/themes_controller.py#L741 
https://github.com/qwc-services/qwc-map-viewer/blob/master/src/qwc2_viewer.py#L1250

Since ?? only falls back for null or undefined, defaultBackgroundLayers was not used.

### DIFF
--- a/actions/theme.js
+++ b/actions/theme.js
@@ -181,7 +181,7 @@ export function setCurrentTheme(theme, themes, preserve = true, initialExtent = 
             printResolutions: theme.printResolutions ?? themes.defaultPrintResolutions,
             printGrid: theme.printGrid ?? themes.defaultPrintGrid,
             searchProviders: theme.searchProviders ?? themes.defaultSearchProviders,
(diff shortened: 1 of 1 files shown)


manisandro:
Though if, for whatever reason, you indeed want a theme to override the default background layers and provide no background layers, specifying an empty array would indeed be the only option?

stephaschu:
Good point. Then there would be no way to disable backgroundLayers when defaultBackgroundLayers are set.
So a better fix would be to adjust qwc-admin-gui to omit the backgroundLayers property instead of storing an empty array.

manisandro:
Yes I think that would be the better fix.

stephaschu:
closing this (https://github.com/qwc-services/qwc-admin-gui/pull/119)

PR #702, opened 2026-07-21
fix isochrone redlining layer

stamogis:
When exporting the isochrone (reachability) routing results as a redlining layer, a layer is created in the layer tree, but no data is visible on the map. This should resolve the issue.


manisandro:
Thanks!

PR #708, opened 2026-07-22
Support value extraction in overlay_intersects expression

dominik-warch:
The overlay_intersects expression only returns true/false and currently does not support additional parameters ([1]). One common use case for us is evaluating a field name from intersecting layers and get back an array of all values to use as default value on creating new features. For example:

```
array_first(overlay_intersects('Flurstücke', "gemarkung"))
```
QGIS Desktop pulls the value from the field `gemarkung` of overlapping polygons and returns the first value (`array_first`). Right now QWC evaluates this expression against the feature being edited instead of the matched features.

This PR adds it's own grammar rule for `overlay_intersects(layer, "field")`. This is a pretty limited scope, the QGIS Desktop implementation allows for arbitrary expressions as second argument which would need a bigger change. Also additional arguments like `filter` or `limit` would be a nice addition later on.

[1] https://docs.qgis.org/3.44/en/docs/user_manual/expressions/functions_list.html#expression-function-geometrygroup-overlay-intersects


manisandro:
Thanks!

PR #712, opened 2026-07-30
Redlining: convert HTML linebreaks to \n

dominik-warch:
QGIS Server encodes line breaks in multiline text fields values as `<br>` or `</br>` tags. The text rendering of OL just puts prints those literal tags. 

This PR replaces `<br>` and `</br>` occurences into `\n` characters for proper linebreaks.


manisandro:
Thanks!

PR #715, opened 2026-08-03
Identify: export numeric-looking attribute values as numbers in XLSX export

dominik-warch:
Attribute values in the XLSX export are exported as strings, so numeric values show up in Excel/LibreCalc with prefixed `'` instead of only the numbers. Aggregating those columns is so not possible without additional work.

This PR adds a small helper function to convert numeric-looking strings cell wise to actual numbers. Data values with leading zeroes should be left as strings to avoid data loss (for example ID etc.)

### DIFF
--- a/components/IdentifyViewer.jsx
+++ b/components/IdentifyViewer.jsx
@@ -38,6 +38,13 @@ import ToggleSwitch from './widgets/ToggleSwitch';
 
 import './style/IdentifyViewer.css';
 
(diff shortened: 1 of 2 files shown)


manisandro:
As this is potentially a generally useful function, I'd call it `isNumeric` and move it to `MiscUtils`.

manisandro:
Thanks!

PR #716, opened 2026-08-04
fix: panoramax layers cleaned after plugin change

Marshlyin:
Bug: When the Panoramax plugin is open and the user doesn't click on the map but opens another plugin instead, the Panoramax custom layers remain visible and don't disappear.

Fix: A new Cleanup() function removes the layers, similarly to the existing onClose() function. It is now called whenever the plugin was active and becomes inactive — i.e. when the user opens another plugin.

### DIFF
--- a/plugins/Panoramax.jsx
+++ b/plugins/Panoramax.jsx
@@ -92,6 +92,8 @@ class Panoramax extends React.Component {
                     this.addRecordingsMVT();
                 }
             }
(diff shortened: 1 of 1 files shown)


manisandro:
Thanks - p.s. no need to call `cleanup()` in `onClose()`, the `this.props.active !== prevProps.active` will be triggered anyway after clearing the task in `onClose()`.

PR #724, opened 2026-08-11
fix: implodeLayers crashes when a layer added via ADD_LAYER_FEATURES has no srcid

Marshlyin:
Bug: creating a new layer feature calling `addLayerFeatures` without an srcid in parameters does not add a fallback uuidv4 value (unlike `addLayer`). Calling implodeLayers then causes React to crash, since the merge check  `prevlayer?.srcid === layer.srcid` resolved in undefined === undefined, resolving in `TypeError: can't access property "sublayers", prevlayer is null`

Fix: adds a default srcid fallback values in `addLayerFeatures` and hardens the check, verifying that prevlayers is not null / srcid is not undefined

### DIFF
--- a/reducers/layers.js
+++ b/reducers/layers.js
@@ -272,7 +272,6 @@ export default function layers(state = defaultState, action) {
             const newLayer = {
                 ...action.layer,
                 id: layerId,
-                srcid: action.layer.srcid ?? uuidv4(),
                 type: 'vector',
                 name: action.layer.name || layerId,
                 features: newFeatures,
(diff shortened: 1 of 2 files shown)


Marshlyin:
According to [copilot](https://github.com/copilot/share/80245226-48c4-8820-b903-1603249500da), the EsLint error is caused by a bad configuration of the `eslint.yml` file, which is weird because the last modification of this file was 6 months ago

manisandro:
eslint is running fine on the master branch, so I believe copilot is wrong.

Marshlyin:
I ran eslint locally and realised that I used the April release on my project and that you fixed that very same bug in May, which resulted in the creation of a duplicate srcid key on my branch. I fixed that part, and it seems to have fixed the error I had on my project. 
The hardened check may not be really relevant, but it still fixes an hypothetical undefined === undefined case.

manisandro:
Hardening the check makes sense, thanks

Issue #534, opened 2025-11-25
support Drag & Drop for importing new layers


4xoc:
It would be great to add additional temporary layers by drag & drop of supported files. The manual steps to import the layers is quite combersome when doing it for multiple files.

manisandro:
Implemented in https://github.com/qgis/qwc2/commit/ded1a2340b3cb92d400b7b1b4d2eb115f959405a

PR #726, opened 2026-08-11
feat: support Drag & Drop for importing new layers

gwerle:
Hey @manisandro! Thank you for maintaining this!

This PR aims for https://github.com/qgis/qwc2/issues/534, to add a drag-and-drop feature and import new layers

My approach was 
- to create `FileImportUtils` using the existing logic for importing new layers
- to develop the new `MapDropImportPlugin`

As this is my first contribution to this project, let me know if something should be changed. Thank you!

image
image


plugins/MapDropImport.jsx
```diff
@@ -0,0 +1,96 @@
+/**
+ * Copyright 2026 Sourcepole AG
```
manisandro: You'll probably want to change this

manisandro:
Thanks! Aside from the one minor comment, looks good!

plugins/MapDropImport.jsx
gwerle: Ah, good catch! I copy-pasted the header from an existing file and forgot to swap that out. I've updated it to my name, but let me know if the project prefers a different attribution format for external contributors.

manisandro:
Thanks!

PR #730, opened 2026-08-19
Identify: add "if" URL parameter to identify features by attribute

### LINKED ISSUE #703: Feature request: open a theme from a URL and identify a feature (or a list of features) by attribute value
## Summary

Add a URL parameter that lets an external application deep-link into QWC2 and, on load, **select and identify one or several features of a given layer by the value(s) of an arbitrary attribute** : highlighting them, zooming to them, and opening the Identify panel with their attributes.

This extends the existing `identifySearchResults` option ([`77d9e85`](https://github.com/qgis/qwc2/commit/77d9e855534fe1b0ae2753017f917ba82c785b0e)) from the single, point-based case to a multi-feature, filter-based one.

## Motivation

External business applications need to hand a user off to a map, passing a reference to one or more objects. That reference is a **business identifier** carried in the external system (a parcel number, an asset id, a tree number...) which:

- may live in **any attribute column**, not necessarily the layer's primary key, so the column has to be specifiable in the URL; and
- is often **not unique**, so a single reference can resolve to several features. Selecting a **list** is therefore the general case, not an edge case.

The expected result: land in the theme with the matching features highlighted, the map zoomed to them, and their attributes visible in the Identify panel.

## Existing approaches (and why they fall short)

QWC2 already exposes three URL-driven ways to deep-link into a theme (see the [Interfacing docs](https://docs.qwc.app/master/topics/Interfacing/)). None cleanly does both **point at a business identifier (or a list)** *and* **select + identify the matching features**.

**1. Generic search — `st` (+ optional `sp` provider).** `SearchBox` runs the theme's search provider(s) on `st`; a single place result auto-selects (`selectPlaceResult`), highlighting + zooming (`showResultGeometry`), and opens Identify with `identifySearchResults: true`. *But:* needs a configured provider, it's a fuzzy text match (not an exact attribute lookup), and only a **single** result triggers it — no list.

**2. Fulltext-search service — `hp` + `hf`/`st`.** The only path built for highlight-by-reference: `FulltextSearch.handleHighlightParameters` queries the data-service, highlights + zooms, and opens Identify with `identifySearchResults: true`. *But:* hard-gated behind `searchServiceUrl` + `dataServiceUrl` (services EMS does not deploy), `hf` is verbose JSON (`[["field","=",value]]`), and the Identify is **point-based** at the bbox centre — fragile for polygons/lines, and wrong for a list.

**3. Theme WMS filter — `f` (MapFilter plugin).** `f` is a JSON blob the `MapFilter` plugin turns into a server-side WMS `FILTER` (`SET_FILTER`). This is the current workaround we use and it works with **no extra service**:

```
?t=ems_test&f=%7B%22__custom%22%3A%5B%7B%22title%22%3A%22Parcelle%22%2C%22layer%22%3A%22parcelles%22%2C%22expr%22%3A%5B%22id_parcellaire%22%2C%22%3D%22%2C%22674820470136%22%5D%7D%5D%7D
```

where the decoded `f` is:

```json
{"__custom": [{"title": "Parcelle", "layer": "parcelles", "expr": ["id_parcellaire", "=", "674820470136"]}]}
```

A list even works out of the box — `"expr": ["id_parcellaire", "IN", ["674820470136", "674820470137"]]` serializes to `IN ( … )`. *But:* it **hides** non-matching features instead of highlighting them, never zooms or opens Identify, needs verbose JSON, and its `IN` serialization leaves values unquoted (so numeric-looking ids only).

## Proposed solution

A dedicated URL parameter, in the spirit of the existing QGIS Web parameters (`t`, `l`, …), of the form:

```
?t=<theme>&<param>=<layer>:<field>:<value1>,<value2>,…
```

On theme load, the viewer would:

1. **parse** the parameter and build an attribute filter of the form `<layer>:"<field>" IN ( <v1> , <v2> , … )` (no JSON in the URL)
2. issue a single **GetFeatureInfo** request carrying that expression as the WMS **`FILTER`** parameter (via QGIS Server), returning every matching feature with geometry and attributes (plus the maptip HTML, if the layer defines one);
3. **populate the `IdentifyViewer`** (multi-feature tree + attribute tables / maptips) and open the panel;
4. **highlight all** matching features (`highlightAllResults`) and zoom to the aggregated bbox.

A single FILTER-based identify covers highlighting and attributes in one path. It is robust (no centroid guessing), does not depend on the search/data services, handles the single-value case as a list of length one, and handles the primary-key case for free (the field can be the PK). The building blocks (`buildFilterRequest`, the `IdentifyViewer` multi-feature tree, `highlightAllResults`) already exist.

## Open questions

- **Parameter name and syntax** : what fits the existing URL-parameter conventions best?
- **List size** : should there be a cap on the number of values, given URL-length and performance/permission considerations, and the server's default `feature_count: 100`?
- **Relation to `hp`/`hf`** : introduce a new, non-verbose parameter and keep `hp`/`hf` as-is, or extend the existing ones?

Happy to implement this and open a PR once there's agreement on the parameter design.

AlexianMasson:
Fixes #703.

Adds an `if` URL parameter which identifies features by attribute on startup: the matching features are queried via WMS GetFeatureInfo, listed in the identify results window, highlighted on the map, and the map is zoomed to their combined extent. This lets an external application deep-link into QWC using its own business identifiers, which is what the generic search (`st`, single fuzzy result), the fulltext search parameters (`hp`/`hf`, point based and requiring a dedicated service) and the theme filter (`f`, which hides non-matching features instead of highlighting them) cannot do today.

## Parameter format

The value is a JSON object mapping a layer to a data service filter expression, which is the same shape as the `filter` of a predefined map filter:

```
if={"<layer>": <data_service_filter_expression>, ...}
```

Examples of valid url with if parameter:
```
?t=qwc_demo&if={"countries":["name","IN",["Germany","Austria","Poland"]]}
?t=qwc_demo&if={"countries":["name","=","France"],"country_names":["name","=","France"]}
?t=qwc_demo&if={"countries":[["pop_est",">",200000000],"or",["name","=","Austria"]]}
```

`<layer>` is the WMS name of a theme layer, optionally prefixed as `<theme_wms_name>#<layer>`, which is required when several WMS layers in the map expose the same layer name and is also what lets an entry be copied straight out of an `f` permalink. Ambiguous unprefixed names are refused with a console warning naming the qualified spellings, rather than silently identifying the wrong layer.

## Limitations

- Like the `f` custom filters, this relies on `wms_name` being present on the theme layer, so it works with themes generated by `qwc-config-generator` and not with the standalone `scripts/themesConfig.js` output.
- The target layer must be visible and queryable when the application starts, and the theme `identifyTool` must be `Identify`.

## Verification

The repository has no test framework, so this was verified against a `qwc-docker` deployment of the demo theme, driving the built viewer headlessly and inspecting the GetFeatureInfo request, the panel contents, the resulting map view and the URL.

### Screenshot

Result for `?t=qwc_demo&if={"countries":["name","IN",["Germany","Austria","Poland"]]}`:
Screenshot_2026-08-19_16-38-04

## Notes

The documentation update is here: [qwc-services.github.io/pull/40](https://github.com/qwc-services/qwc-services.github.io/pull/40)

## Sponsor

### DIFF
--- a/plugins/Identify.jsx
+++ b/plugins/Identify.jsx
@@ -173,8 +173,7 @@ class Identify extends React.Component {
         exitTaskOnResultsClose: null,
         filterGeom: null,
         filterGeomModifiers: {},
-        importErrors: null,
-        pendingIdentifyFilter: null
+        importErrors: null
     };
     constructor(props) {
         super(props);
@@ -183,6 +182,7 @@ class Identify extends React.Component {
         this.fileinput.accept = "application/json";
         this.fileinput.addEventListener("change", this.fileSelected);
         this.viewerRef = null;
+        this.pendingIdentifyFilter = null;
     }
     componentDidUpdate(prevProps, prevState) {
         if (this.props.theme && !prevProps.theme) {
@@ -195,7 +195,7 @@ class Identify extends React.Component {
             } else if (startupParams.if) {
                 // Handled even when Identify is not the current identify tool.
                 // Deferred until the map filter is applied, see identifyFeaturesPending
-                this.setState({pendingIdentifyFilter: startupParams.if});
+                this.pendingIdentifyFilter = startupParams.if;
                 UrlParams.updateParams({if: undefined});
             } else if (this.props.enabled && this.props.startupState.identifyresultstate) {
                 this.deserializeResults(
@@ -210,7 +210,7 @@ class Identify extends React.Component {
                 this.clearResults();
             }
         }
-        if (this.state.pendingIdentifyFilter) {
+        if (this.pendingIdentifyFilter) {
             this.identifyFeaturesPending();
         }
         if (this.props.enabled) {
@@ -376,8 +376,8 @@ class Identify extends React.Component {
         if (awaitMapFilter && this.props.layerFilter?.filterParams === null) {
             return;
         }
-        const identifyFilter = this.state.pendingIdentifyFilter;
-        this.setState({pendingIdentifyFilter: null});
+        const identifyFilter = this.pendingIdentifyFilter;
+        this.pendingIdentifyFilter = null;
         this.identifyFeatures(identifyFilter);
     };
     identifyFeatures = (identifyFilter) => {
--- a/plugins/Identify.jsx
+++ b/plugins/Identify.jsx
@@ -274,8 +274,8 @@ class Identify extends React.Component {
             const appendResultsByDefault = this.props.appendResultsByDefault === true || (this.props.taskEnabled && this.props.appendResultsByDefault === "task");
             const identifyResults = this.props.click.modifiers.ctrl !== true && !appendResultsByDefault ? {} : state.identifyResults ?? {};
 
-            const params = {...this.props.params};
             const resultDisplayMode = this.props.taskEnabled ? (this.props.taskResultDisplayMode ?? this.props.resultDisplayMode) : this.props.resultDisplayMode;
+            const params = this.buildIdentifyParams();
             let queryableLayers = [];
             queryableLayers = IdentifyUtils.getQueryLayers(this.props.layers, this.props.map);
             queryableLayers.forEach(l => {
@@ -323,12 +323,8 @@ class Identify extends React.Component {
         const appendResultsByDefault = this.props.appendResultsByDefault === true || (this.props.taskEnabled && this.props.appendResultsByDefault === "task");
 
         let pendingRequests = 0;
-        const params = {...this.props.params};
         const resultDisplayMode = this.props.taskEnabled ? (this.props.taskResultDisplayMode ?? this.props.resultDisplayMode) : this.props.resultDisplayMode;
-        if (this.props.params.region_feature_count) {
-            params.feature_count = this.props.params.region_feature_count;
-            delete params.region_feature_count;
-        }
+        const params = this.buildIdentifyParams("region_feature_count");
         const requestFilterGeom = this.getRequestFilterGeomWkt(filterGeom);
         // Querying without the filter geometry would match the whole layer
         if (requestFilterGeom) {
@@ -347,6 +343,16 @@ class Identify extends React.Component {
             const identifyResults = state.filterGeomModifiers.ctrl !== true && !appendResultsByDefault ? {} : state.identifyResults ?? {};
             return {identifyResults: identifyResults, pendingRequests: pendingRequests, currentResultDisplayMode: resultDisplayMode};
         });
+    };
+    buildIdentifyParams = (featureCountParam = null) => {
+        const params = {...this.props.params};
+        if (featureCountParam && params[featureCountParam]) {
+            params.feature_count = params[featureCountParam];
+        }
+        // These are internal params, not GetFeatureInfo params
+        delete params.region_feature_count;
+        delete params.radius_feature_count;
+        return params;
     };
     getRequestFilterGeomWkt = (identifyGeom) => {
         const sourceFilterGeom = this.props.layerFilterGeom;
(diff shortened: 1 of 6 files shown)


plugins/Identify.jsx
```diff
@@ -173,16 +179,22 @@ class Identify extends React.Component {
...
             const haveIc = ["1", "true"].includes((startupParams.ic || "").toLowerCase());
             const c = (startupParams.c || "").split(/[;,]/g).map(x => parseFloat(x) || 0);
-            if (haveIc && c.length === 2) {
+            if (this.props.enabled && haveIc && c.length === 2) {
                 const mapCrs = this.props.theme.mapCrs;
                 this.identifyPoint(CoordinatesUtils.reproject(c, startupParams.crs || mapCrs, mapCrs));
-            } else if (this.props.startupState.identifyresultstate) {
+            } else if (startupParams.if) {
+                // Handled even when Identify is not the current identify tool.
+                // Deferred until the map filter is applied, see identifyFeaturesPending
```
manisandro: Can you elaborate on this? Is the issue that you want to ensure a filter set in a separate `f` URL parameter is applied to `if`?

AlexianMasson: Yes, exactly that. `identifyFeatures` seeds the per-layer filters from `layers.filter.filterParams` and joins the `if` expression onto them, so MapFilter must have applied the startup filter first. That is asynchronous:

MapFilter schedules `applyFilter` behind a 500 ms debounce (`plugins/MapFilter.jsx:146-147`) before `setFilter` reaches the store, while the Identify startup branch runs as soon as the theme is available, when `filterParams` is still `null`. Identifying at that point would send a FILTER carrying only the `if`
expression, so the results would include features that `f` hides.

The wait only applies while `f` is set and MapFilter is configured, so a plain `if=` URL identifies immediately.

I will move the explanation onto `identifyFeaturesPending` and mention the debounce there.

manisandro: Two things I noticed while reviewing:

- Handling the `if` parameter even if the Identify tool is not enabled can have an unexpected effect if the theme defines a startupTask: closing the window may call `setCurrentTask(null)` and terminate the running task. You should set `exitTaskOnResultsClose: false` in the state when the results are displayed triggered by an `if` URL parameter
- If the `f` contains an invalid filter, the identify request will never be processed. But I guess this can be accepted.

AlexianMasson: > * Handling the `if` parameter even if the Identify tool is not enabled can have an unexpected effect if the theme defines a startupTask: closing the window may call `setCurrentTask(null)` and terminate the running task. You should set `exitTaskOnResultsClose: false` in the state when the results are displayed triggered by an `if` URL parameter

Yes indeed, nice catch. I fixed it with a guard on the task instead of a state flag. FeatureForm already has this kidn of guard, so I better use the same one. Here : https://github.com/qgis/qwc2/pull/730/commits/ac2aa81bb258263a14100180bd71ef2653a69442

Note that `exitTaskOnResultsClose: false` in the state alone doesn't fix it: `onWindowClose` uses `||`, so it loses against a plugin config setting the prop to true, which is the only case where the task gets killed. I checked and the startup task still dies. It would need `??` instead of `||`. Tell me if you'd rather have the state flag.

> * If the `f` contains an invalid filter, the identify request will never be processed. But I guess this can be accepted.

I  checked and the request does go through (with an url like : `http://127.0.0.1:8081/?t=qwc_demo&if={"countries":["name","=","France"]}&f={"unclosed`)

manisandro:
Thanks!

AlexianMasson:
Thanks for the reviewing work !

manisandro:
@AlexianMasson I had to refactor Identify a bit (https://github.com/qgis/qwc2/commit/c9d91755a2c2600a8025d32f007da89db7cc92ce) to allow adding the new `FeatureSelection` plugin. Could you please verify that your use-case with the `if` parameter still works as desired?

AlexianMasson:
@manisandro All good, working fine :+1:

manisandro:
Thanks!

PR #732, opened 2026-08-24
Create a generalized search widget

mki-c2c:
The search widget currently only allows text content in the search results.

This PR created a generalised parent which can either be used as the legacy search widget or can be filled with custom component as result items.
This can be used to implement rich result display with embedded images and links.

### DIFF
--- a/components/widgets/SearchWidget.jsx
+++ b/components/widgets/SearchWidget.jsx
@@ -16,16 +16,50 @@ import {v4 as uuidv4} from 'uuid';
 
 import LocaleUtils from '../../utils/LocaleUtils';
 import {SearchResultType} from '../../utils/SearchProviders';
-import VectorLayerUtils from '../../utils/VectorLayerUtils';
-import Icon from '../Icon';
-import InputContainer from './InputContainer';
-import PopupMenu from './PopupMenu';
-import Spinner from './Spinner';
+import SearchWidgetBase from './SearchWidgetBase';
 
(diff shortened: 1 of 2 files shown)


manisandro:
Hi @mki-c2c 

I've pushed a less invasive approach here: https://github.com/qgis/qwc2/commit/fef4b34eb39c57b3f89b6e84b6aedb802974ab87

Let me know if you have any issues.

mki-c2c:
looks great, thanks.

I'll try using it in the custom child widget

mki-c2c:
hello @manisandro 
I noticed abstraction of the click handler was also needed:
https://github.com/mki-c2c/qwc2/pull/3
could you please also take that one into account ?

manisandro:
Can't you just pass your custom onClick callback in the DOM fragment returned by the custom `renderItem`?

mki-c2c:
yes I found a workaround, but the geometry part in searchWidget is lost:
https://github.com/qgis/qwc2/blob/fef4b34eb39c57b3f89b6e84b6aedb802974ab87/components/widgets/SearchWidget.jsx#L169

I'm not sure when this is useful, but without a onclick prop, this code becomes inaccessible and would have to be rewritten if needed.

Let's keep it like this for the time being :-)

manisandro:
I see - In case there is indeed no harm in passing the default onClick handler as an additional parameter to the render function, as you propose.

PR #733, opened 2026-08-26
ImportLayer: support importing DXF files

dominik-warch:
Adds DXF to the local file import

Using the `dxf` package from npm DXF entities are converted to polylines and added as line features. This way the imported DXF reuses the existing vector layer code path for reprojection etc.

Since DXF itself has no CRS it just puts the DXF into the current map CRS but a dropdown menu allows for selecting a different CRS.

A bigger issue where some weird encoding problems. After some research it looks like DXF pre 2007 does not use UFT-8 so I added $DWGCODEPAGE to decode those DXF files accordingly.

Currently there is no support for TEXT/MTEXT entitites, which are dropped entirely. I hope to add them in a later commit if this PR moves forward. But this 'basic' version is already enought for 80 % of our use cases.


manisandro:
Thanks!! Looks good!

PR #734, opened 2026-08-26
Update feature status based on data presence

mki-c2c:
the 'new' status is important, otherwise the relation values do not get saved correctly, cf. https://github.com/mki-c2c/qwc2/pull/1/changes

### DIFF
--- a/components/AttributeForm.jsx
+++ b/components/AttributeForm.jsx
@@ -239,7 +239,7 @@ class AttributeForm extends React.Component {
             type: "Feature",
             properties: {...initialProperties}
         }, this.props.iface, this.props.editContext.mapPrefix, this.props.map.projection, newRelFeature => {
-            newRelFeature.__status__ = data ? "new" : "empty"
+            newRelFeature.__status__ = initialProperties ? "new" : "empty"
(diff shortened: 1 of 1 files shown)


components/AttributeForm.jsx
```diff
@@ -239,7 +239,7 @@ class AttributeForm extends React.Component {
             type: "Feature",
             properties: {...initialProperties}
         }, this.props.iface, this.props.editContext.mapPrefix, this.props.map.projection, newRelFeature => {
-            newRelFeature.__status__ = "empty";
+            newRelFeature.__status__ = data ? "new" : "empty"
```
manisandro: `data` -> `initialProperties`?

mki-c2c: of course, sorry!

mki-c2c: fixed

manisandro:
Looking at this more closely, I don't like the new vs empty thing. I'm working to refactor the nrel code to avoid it.

manisandro:
=> https://github.com/qgis/qwc2/commit/613692bf830cd306f627d7755b364c965d69aff4

PR #735, opened 2026-08-27
DxfUtils, MapExport: fix block/linetype/layer name and handle collisions when merging DXF exports

### LINKED ISSUE #686: Invalid DXF export after adding layers from another theme
Exporting the map as DXF works correct when I only export layers from the current theme. As soon as I add layers from a different theme, the exported DXF is broken/invalid. Some software like BricsCAD can open them but others either can't open the file at all or show missing or empty layers.

dominik-warch:
Fixes #686 - Atleast tried the import in QGIS, FreeCAD, BricsCAD and AutoCAD. The most robust variant would be to outright ban DXF globally and get rid of this code *fingers crossed*

I dug deeper into this issue than anticipated, fixing one thing showed some new things popping up in one CAD tool but not in another one. What did I change...

1. Block name collision: QGIS assigns block names (`symbolLayerN`) starting from 0 in every document. During merge, all non shared entries from document 2 on onward get renamed with a suffix `_m$` and `INSERT` references get rewritten via `blockNameMapping` which runs recursively into all nested blocks.
2. Handle collision: Handles and owner pointers were only reassigned for top level entties, `remapHandles` now runs recursivley into all block children and `tailValues`.
3. LTYPE table was not merged at all: QGIS names linetypes using the same schema but indepently of the block names. This LTYPE table is now merged the same way.
4. LAYER duplicates: Existing layer names are now deduplicated and not blindly appended to the first document.
5. `$DWGCODEPAGE` is now normalized. QGIS Server writes `8859_1` instead of the DXF spec convetion `ANSI_1252`. A new function rewrites the header value to the corred for based on the encoding actually used for the output. This was at least a warning in some CAD tools, maybe not breaking though.
6. Encoding now dynamic based on `detectDxfEncoding/decodeDxf` instead of the hardcodesd `iso-8859-1`. This did not change any behaviour in my case but sets the correct codes for UTF-8 documents (AC1021+)

I still have some issues regarding the Z-layer ordering but I am not sure if this is a problem from the layer tree itself or some weird QGIS DXF export issues or also part of the merge.

### DIFF
--- a/plugins/MapExport.jsx
+++ b/plugins/MapExport.jsx
@@ -26,7 +26,7 @@ import NumberInput from '../components/widgets/NumberInput';
 import Spinner from '../components/widgets/Spinner';
 import ConfigUtils from '../utils/ConfigUtils';
 import CoordinatesUtils from '../utils/CoordinatesUtils';
-import {explodeDxf, implodeDxf, mergeDxf} from '../utils/DxfUtils';
+import {decodeDxf, detectDxfEncoding, explodeDxf, implodeDxf, mergeDxf, setDwgCodepage} from '../utils/DxfUtils';
 import LayerUtils from '../utils/LayerUtils';
 import LocaleUtils from '../utils/LocaleUtils';
 import MapUtils from '../utils/MapUtils';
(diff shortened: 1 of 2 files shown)


manisandro:
I can only agree with `The most robust variant would be to outright ban DXF globally and get rid of this code` :laughing: Thanks!!

PR #737, opened 2026-09-01
fix IdentifyViewer: add default for innerRef prop

dominik-warch:
`innerRef` is optional but is called everytime in `componentDidMount` and `componentWillUnmoun`. The prop is introduced in 3272dec1 but only in `Identify.jsx`. FeatureSearch results crash with `TypeError: this.props.innerRef is not a function` because there the prop is not set.

This PR sets `innerRef` in `defaultProps` so code that utilizies the `IdentifyViewer` component without setting innerRef do not crash.

### DIFF
--- a/components/AttributeForm.jsx
+++ b/components/AttributeForm.jsx
@@ -232,37 +232,51 @@ class AttributeForm extends React.Component {
             }
         }
     };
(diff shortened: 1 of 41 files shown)


manisandro:
Thanks!

PR #738, opened 2026-09-01
PrintSelection: allow changing the aspect ratio of freeform selections

dominik-warch:
If the MapExport is configured without a page size, the freeform export frame/printing selection can only be changed with the fixed ratio it is initially drawn. I think this behavior is correct for fixed page sizes where the ratio comes from the page itself but not for freeform extents.

This PR changes the default behavior for printing selection without configured page sizes to be completelty freeform while preserving the aspect ratio while holding the Shift key (behavior in other software like CAD etc.).

### DIFF
--- a/components/AttributeForm.jsx
+++ b/components/AttributeForm.jsx
@@ -232,37 +232,51 @@ class AttributeForm extends React.Component {
             }
         }
     };
-    addRelationRecord = (table, initialProperties = {}) => {
-        const newRelationValues = {...this.props.editContext.feature.relationValues};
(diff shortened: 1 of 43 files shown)


manisandro:
Does this affect the print frame in the Print plugin?

dominik-warch:
No, since the Print plugin always passes `fixedFrame`. The new default behavior is gated behind `fixedFrame === null`, so PrintPlugin or MapExport with configured page sizes only allow for the current aspect ratio preserving scaling.

manisandro:
Thanks!

PR #739, opened 2026-09-01
MapUtils: derive getZoomForExtent from the exact zoom level so the extent always fits

AlexianMasson:
`MapUtils.getZoomForExtent` documents itself as returning "the zoom level fitting the extent", but with `allowFractionalZoom` disabled (the default) it can return a level which is zoomed in past the requested extent, cutting off part of it.

## Why it matters

Reported by users of the `if` URL parameter added in #730, which zooms to the combined extent of the identified features. When the extent falls in the bad range, the map lands one level too deep and the features the URL is about are outside the viewport, even though the identify results window lists them.

The screenshots below are the same URL against the same theme, before and after the fix. Two trees 123 m apart are identified by attribute. Before, the map goes to 1:500 and both features are off screen. After, it goes to 1:1000 and both are visible:

| | |
|---|---|
| before | zoom-fit-before |
| after | zoom-fit-after |

## Other affected features

Some other features seems to compensate this issue by dropping from one level of zoom and I updated since it's not needed anymore. In my opinion, it can only improve the fitting of the zoom extent but maybe I'm missing some edge cases... https://github.com/qgis/qwc2/pull/739/commits/93b01522ceea66b7005d03a62ceb06f3b5729d9f

## Notes

- No behaviour change when `allowFractionalZoom` is enabled.
- `computeFractionalZoom` is a straight extraction, not a rewrite, so `computeZoom` returns exactly what it returned before.

### DIFF
--- a/components/IdentifyViewer.jsx
+++ b/components/IdentifyViewer.jsx
@@ -971,7 +971,7 @@ class IdentifyViewer extends React.Component {
         let zoom = 0;
         const maxZoom = MapUtils.computeZoom(this.props.map.scales, this.props.theme.minSearchScaleDenom || 1000);
         if (result.bbox[0] !== result.bbox[2] && result.bbox[1] !== result.bbox[3]) {
-            zoom = Math.max(0, MapUtils.getZoomForExtent(result.bbox, this.props.map.resolutions, this.props.map.size, 0, maxZoom + 1) - 1);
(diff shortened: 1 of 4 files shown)


AlexianMasson:
Hello @manisandro, I tried to do a quick fix. Tell me what you think about it. If it's more complicated than that, I can close this PR and open a dedicated issue to talk about it and about all the affected features.

manisandro:
Looks good to me, thanks!

PR #751, opened 2026-09-04
Redlining: add square and star symbols for point geometries

### LINKED ISSUE #714: Feature request: predefined shapes (circle, square, star) for redlining points
Add a choice of predefined symbol: **circle, square or star** for point geometries in the redlining tool, selected from a flyout on the Point draw button in the same way the polygon shapes work today.

The geometry stays a `Point`; only the symbol used to render it differs.

## Motivation

Points drawn with the redlining tool always render as a circle. Polygons already offer four shapes on their draw button (Circle/Ellipse/Square/Box), but points have no equivalent.

Different symbols would make categories of point annotation easier to tell apart at a glance, and they stay distinguishable on a greyscale print, where colour is less reliable.

## Proposed solution

A `pointShape` property on the feature's `styleOptions`, with values `circle` (the current behaviour and the default), `square` and `star`.

- **UI** : the single Point entry in `drawButtons` becomes a flyout group, exactly as the Polygon one is. The existing Point entry keeps its current icon and label, so the only visible change to the toolbar is the chevron. The two added entries are gated by `hiddenTools`, like the polygon shapes, so a deployment that only wants plain points can hide them.
- **Screen** : the two new symbols are drawn with `ol.style.RegularShape` in `defaultStyle` (circle keeps `ol.style.Circle`). Each needs its own radius factor, since `RegularShape`'s radius is a circumradius, passing the existing `circleRadius` unadjusted would come out at noticeably different widths per symbol for the same size setting.
- **Print** : `pointShape` selects the `<se:WellKnownName>` of the point symbolizer in the generated SLD instead of the hardcoded `circle`. The three values are standard well-known mark names, so no mapping table is needed. Map export goes through the same highlight params, so it is covered by the same change. I have confirmed a QGIS Server renders the three marks via GetMap; I would validate the full GetPrint path as part of the implementation.
- **Persistence** : nothing to add: `styleOptions` is already serialised as a top-level field in the exported GeoJSON and read back on import.
- **Compatibility** : features drawn before the option existed carry no `pointShape` and fall back to a circle, as does an unrecognised value, so existing layers and saved files are unaffected.

Image

## Open questions

- **What to call the property.** The chosen symbol is stored per feature and written into exported GeoJSON, so renaming it after release would leave points in already-saved files falling back to circles. I would use `pointShape`; say if you would rather it were something else, e.g. `pointSymbol`.
- **The shape set.** I'd start with these three, since that is the actual request, the server also renders `triangle`, `cross` and `x` if you would rather have them up front.

I would be glad to implement this and open a PR once the approach is agreed.

ronitjadhav:
Fixes #714

Adds a choice of predefined symbol (circle, square or star) for point geometries in the redlining tool, selected from a flyout on the Point draw button in the same way the polygon shapes work today. The geometry stays a `Point`; only the symbol used to render it differs.

## How it works

The symbol is stored per feature as `pointShape` in `styleOptions`, with values `circle` (the default and the previous behaviour), `square` and `star`.

* **Screen**: `square` and `star` use `ol.style.RegularShape`, `circle` keeps `ol.style.Circle`. Each carries its own radius factor, since `RegularShape`'s radius is a circumradius and passing `circleRadius` unadjusted would give a different width per symbol at the same size setting.
* **Print**: `pointShape` selects the `<se:WellKnownName>` of the point symbolizer instead of the hardcoded `circle`. The three values are standard well-known mark names, so no mapping table is needed, and the value is validated against the shape table before reaching the XML since it can arrive from an imported file. Map export and the height profile use the same highlight params.

The point symbolizer now also emits `stroke-linejoin: round`, which the same function already sets for the line and polygon symbolizers.

The two added entries are gated by `hiddenTools` like the polygon shapes (`"hiddenTools": ["PointSquare", "PointStar"]`). Features drawn before this option existed carry no `pointShape` and fall back to a circle, as does an unrecognised value, so existing layers and saved files are unaffected. The existing Point entry keeps its current icon and label.

## Verification

There is no test framework, so this was exercised manually against the demo theme:

1. Drawing each symbol, and that picking a shape restyles neither the point just drawn nor an existing one on selection.
2. Hiding each new entry with `hiddenTools`.
3. A GeoJSON export and reimport round trip.

### Screenshot

image

## Notes

Documentation update: https://github.com/qwc-services/qwc-services.github.io/pull/41

## Sponsor


manisandro:
LGTM, thanks!

PR #764, opened 2026-09-17
Redlining: add a triangle point symbol

ronitjadhav:
Follow-up to #751, which added circle, square and star point symbols. This adds the triangle that was missed there.

* One entry in the shape table, plus the matching flyout button, gated by `hiddenTools` as `PointTriangle`.
* The radius factor `2 / Math.sqrt(3)` converts `circleRadius` into the `RegularShape` circumradius, so the triangle comes out the same width as the other symbols at a given size. Same convention as `square` and `star`.
* Printing needs no change: `createSld` already validates `pointShape` against the shape table and emits it as the `<se:WellKnownName>`.

## Verification

Verified by drawing all four symbols against the demo theme and rendering the generated SLD through a QGIS Server. 

### Screenshot
image

## Notes

Documentation update: [qwc-services.github.io/pull/41](https://github.com/qwc-services/qwc-services.github.io/pull/41), amended to cover the triangle.

## Sponsor


manisandro:
Thanks!

Issue #514, opened 2025-10-28
Print template with differents layer styles


cazitouni:
Hello @manisandro,

I’m experiencing an issue when using the print functionality with templates that contain more than one map.

For example, when a template includes two maps with different layer styles, like this:

<img width="694" height="522" alt="Image" src="https://github.com/user-attachments/assets/d6fe204d-9dd0-4049-8416-3ec17f63fc27" />

Only the last map is actually printed when using the QWC2 print function together with the qwc-print-service:

<img width="915" height="722" alt="Image" src="https://github.com/user-attachments/assets/d25d84e3-8e1a-4245-a1c1-21d57e7cfca7" />

Did I miss something in the configuration, or is this the intended behavior?
If it’s a bug, I’d be happy to help investigate further.

Best regards,
Clement

manisandro:
Hi @cazitouni 

This is the currently implemented behaviour, the config generator will just look for one layout map item [1]. Potentially this could be extended to support multiple map items, and the QWC client will then also need to be adapted, to either send the selected extend for all maps in the GetPrint request, or more complicated, allow multiple map selections.

[1] https://github.com/qwc-services/qwc-config-generator/blob/master/src/config_generator/qgs_reader.py#L183

cazitouni:
Ok @manisandro,

Thanks for the quick reply!
I’ll try to take a look at it when I find some time.

Maybe we could keep this ticket open as an enhancement request?

Best regards,
Clément

AlexianMasson:
Hi @manisandro,

I'd like to pick up this feature on behalf and funded by the Eurométropole de Strasbourg (EMS), in coordination with @cazitouni .

I dug through QGIS Server, the config-generator and QWC2 to investigate about this issue. Indeed, QGIS Server already renders N maps correctly if the request carries per-map parameters. It's a missing feature in the QWC chain (config-generator + QWC2). Here is a few key findings from my investigation :

[...]

PR #766, opened 2026-09-17
Print: resolve a layout map's theme the way the desktop does

AlexianMasson:
Three defects in one code path, which together mean a layout map that follows a map theme never printed the way QGIS desktop renders it. Independent of #514, though the multi map work builds on the same helper.

1. **The path never ran.** Preset entries were read as `options.visible`, but the config generator writes `{checked, style}` and `LayerUtils` reads `checked` everywhere else. So on current master, any layout whose map follows a map theme prints with `LAYERS=` empty, that is, blank.
2. **Wrong layer order.** The desktop resolves a theme through `mapThemeVisibleLayers`, which orders by the project layer order, not by the order the theme was captured in. The client emitted capture order, and did not reverse to the bottom first order the server expects, the way `LayerUtils.collectPrintParams` does. The entries are now sorted against the client's own theme layer tree, which is the project order unless the user has reordered layers, and then reversed.
3. **Checked group nodes were requested as layers.** They live in the same preset dict with no `style` key, so the group name went into `LAYERS` beside its own member layers, and `undefined` went into the positional `STYLES` list.

Also, the block overwrote the global `LAYERS` while leaving `OPACITIES` and `COLORS` in place, describing a different layer list of a different length, which the server applies positionally. Both are now dropped on this path: the theme's styles carry the opacity, as on the desktop.

Needs https://github.com/qwc-services/qwc-config-generator/pull/131 to land first, otherwise this makes a second, separate bug live. Worth backporting together.

---

### Sponsor


manisandro:
Thanks!

PR #767, opened 2026-09-21
Print: render every map of a multi-map print layout

### LINKED ISSUE #514: Print template with differents layer styles

AlexianMasson:
Fixes #514. Needs [qwc-services/qwc-config-generator#132](https://github.com/qwc-services/qwc-config-generator/pull/132), which reports the extra maps as `fixedMaps`. Without it this PR is inert. Builds on #766.

QGIS Server removes any non atlas driven layout map the request carries no `mapN:EXTENT` for, and the print form only ever sent parameters for one map. So an inset, a locator or a second thematic map was deleted from the sheet.

### The contract

A frozen map prints as QGIS desktop renders it. The interactive map prints what the user sees. Everything below follows from that.

* Send `mapN:EXTENT` for every entry of `fixedMaps`.
* Send it unchanged when the map item carries its own CRS: the server keeps that CRS and assigns the extent to it with no transform, and the designer reprojects the saved extent when the CRS is changed, so the numbers are already right.
* Otherwise reproject, sampling along the edges rather than only the corners, then restore the aspect of the authored on page rectangle. `setExtent` keeps the item's width and recomputes its height from the extent's aspect, so an uncorrected extent silently reshapes the item and clips it. An inset authored at 0.690 reprojects to 0.722.
* Skip a map whose extent cannot be reprojected, with a console warning, rather than sending a degenerate extent.
* Resolve a map theme into explicit `mapN:LAYERS` and `mapN:STYLES`. The server back fills the layer set from the global `LAYERS` before it tests the frozen state, so such a map keeps the theme's styles but not its layer set.
* Resend the authored `mapN:GRID_INTERVAL_X` and `_Y`, which the server otherwise resets to zero, wiping an authored graticule.

### Verification

A Lambert 93 project with 12 print layouts, printed end to end through the real client and QGIS Server, on master and on this branch, in two map CRS, and compared against the same layouts exported from QGIS desktop with `QgsLayoutExporter`.

Eleven of the twelve now match the desktop export within 0.16% to 0.47% of pixels, down from 3.7% to 43%, the remainder being antialiasing and the server to desktop version gap. For a single map layout the emitted request is identical before and after, in both map CRS.

Four of those layouts are cases master could not print at all: an inset carrying its own map CRS, an atlas sheet with a frozen inset, a theme with two overlapping opaque layers, and a theme with the group node itself checked.

### Known limitation

A layout with more than one non-frozen map keeps the first as the interactive one; the server removes the others, as it did before. The config generator now warns for each. Ticking *Follow map theme* without choosing a theme lands here, because QGIS itself then treats the map as an ordinary map. The mirror case, a multi map layout with no unfrozen map at all, keeps the first map as the interactive one and ignores its freeze; the config generator warns for that too.

Happy to cover it here, or in another PR if you prefer.

---

### Sponsor


manisandro:
Thanks!

